PR #1, opened 2020-06-06
Svg removal

atomiccupcake:
Switching styling out of js/svg logic and into css


liwords-ui/src/gameroom/board_panel.tsx
```diff
@@ -236,13 +229,14 @@ export const BoardPanel = (props: Props) => {
...
       </div>
-      <div style={{ marginTop: 30 }}>
+      <div>
         <GameControls
-          onExchange={exchangeTiles}
-          onPass={passTurn}
-          onChallenge={challengePlay}
-          onCommit={commitPlay}
-          currentRack={props.currentRack}
+            onRecall={recallTiles}
```
domino14, line 234: Minor thing here -- make sure your Prettier runs automatically so that all code is formatted the same (this should be two spaces)

domino14:
Screen Shot 2020-06-06 at 4 24 20 PM

Overall this looks like a huge improvement on the SVG-based code, I especially like moving all the styles over to css and using proper class names for everything, thank you for this cleanup.

I noticed a couple of issues with resizing, see image above.

- The tiles don't seem to be centered on their squares in the board at the size of my laptop (when I shrunk or grew the window, the tiles also seem to resize a bit funny and get more out of center, even moving to other squares).
- The tiles on the rack are not the right size (not sure if you touched this with this PR)

atomiccupcake:
> Screen Shot 2020-06-06 at 4 24 20 PM
> 
> Overall this looks like a huge improvement on the SVG-based code, I especially like moving all the styles over to css and using proper class names for everything, thank you for this cleanup.
> 
> I noticed a couple of issues with resizing, see image above.
> 
> * The tiles don't seem to be centered on their squares in the board at the size of my laptop (when I shrunk or grew the window, the tiles also seem to resize a bit funny and get more out of center, even moving to other squares).
> * The tiles on the rack are not the right size (not sure if you touched this with this PR)

Yeah. Wasn't rendering that way before the merge, so no idea what it's doing now, but I have some ideas and am pretty sure it will be easier to straighten out once I have the backend running. We can get those changes added to this pr before merging in.

PR #2, opened 2020-06-06
[WIP] Using reducers for state management

domino14:


liwords-ui/src/gameroom/table.tsx
```diff
@@ -79,8 +75,19 @@ export const Table = (props: Props) => {
...
+  // the player on turn.
+  let rack;
+  const us = gameContext.players.find((p) => p.nickname === props.username);
+  if (us) {
+    rack = us.currentRack;
+  } else {
+    rack = gameContext.players.find((p) => p.onturn)?.currentRack || '';
+  }
+
+  console.log('in table', gameContext.players);
```
jvc56, line 90: Why are we only displaying one of the racks? Why not both? This is one of the things that annoys me about ISC.

domino14: I agree, I also want to display both racks, but Conrad wetoed it as he was concerned it made it too easy to cheat by having a watcher on the table. We could discuss, I think this would be better as well.

PR #16, opened 2020-07-05
Dragdrop v1

atomiccupcake:
Fixes a lot of bugs with sizing, changes styles to Conrad's newests, drag and drop allows rearranging your rack (with mouse only)

### DIFF
--- a/liwords-ui/src/gameroom/scorecard.tsx
+++ b/liwords-ui/src/gameroom/scorecard.tsx
@@ -1,4 +1,4 @@
-import React, { useMemo, useRef, useEffect } from 'react';
+import React, { useMemo, useRef } from 'react';
 import { Card } from 'antd';
 import {
   GameTurn,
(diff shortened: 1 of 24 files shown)


liwords-ui/src/gameroom/scorecard.tsx
```diff
@@ -130,11 +130,8 @@ const Turn = (props: turnProps) => {
   );
 };
 
-export const ScoreCard = (props: Props) => {
+export const ScoreCard = React.memo((props: Props) => {
   const el = useRef<HTMLDivElement>(null);
-  useEffect(() => {
-    el.current?.scrollIntoView({ block: 'end', behavior: 'smooth' });
```
domino14, line 136: How do we get the scorecard content to scroll into view (without scrolling the entire window)? is it possible?

liwords-ui/src/gameroom/tile.tsx
```diff
@@ -37,10 +36,12 @@ const TILE_STYLES: { [name: string]: TileStyle } = {
 };
 
 type TileLetterProps = {
-  rune: string;
+  rune: string,
```
domino14, line 39: this should be a semicolon

liwords-ui/src/gameroom/scorecard.tsx
atomiccupcake: Ooh! That's why that was there. Yeah, I can do that.

PR #17, opened 2020-07-06
some enhancemnets

domino14:


domino14:
This still has a few bugs, but is event-based rather than turn-based now. This should result in easier to maintain and less buggy code.

PR #20, opened 2020-07-20
Drag drop plays

atomiccupcake:
Commiting this even though there are two defects I am aware of: there is a bug with designating blanks if you have 2. Also it doesn't currently handle dragging back to empty space in the rack just to tiles , which is annoying. Both I'll fix in next pr, since this is pretty useful despite that, unless you object

Also, not sure prettier is working right on this machine... let me know if it isn't.


domino14:
Those bugs are fine, we can merge this PR before fixing them. I think prettier isn’t working right though from seeing some of the changes it has made. Is it pointing at the right .prettierrc file (in the liwords-ui directory)?

PR #26, opened 2020-08-03
Autopairing

jvc56:
It's kind of a mess but it works (I think).


domino14:
That autopair internal file is impressive :O

PR #49, opened 2020-08-30
Stat list separation

jvc56:
It's a mess.


pkg/gameplay/game_test.go
```diff
@@ -118,11 +130,13 @@ func TestComputeGameStats(t *testing.T) {
 	is.NoErr(err)
 
 	ctx := context.WithValue(context.Background(), ConfigCtxKey("config"), &DefaultConfig)
-	stats, err := computeGameStats(ctx, hist, gameReq, variantKey(req), ustore)
+	// Fix this once list separation is complete
```
domino14, line 133: does this comment still apply? what needs to be fixed?

jvc56: We need to implement AddListItem and GetListItems

pkg/gameplay/game.go
```diff
@@ -45,8 +46,22 @@ type GameStore interface {
...
+}
+
+func (lss DummyListStatStore) GetListItems(statType int, gameIds []string, playerId string) ([]*entity.ListItem, error) {
+	return []*entity.ListItem{}, nil
+}
+
 type ConfigCtxKey string
 
+var listStatStore = DummyListStatStore{}
+
```
domino14, line 64: I don't think this or the DummyListStatStore need to be defined here. Instead some function here (perhaps PlayMove) would take in a ListStatStore interface. The current signature looks like this for example:

```
func PlayMove(ctx context.Context, gameStore GameStore, userStore user.Store, userID string,
	cge *pb.ClientGameplayEvent) error {
```

Note that the GameStore and user.Store types are interfaces and not necessarily structs. Additionally, PlayMove perhaps could take in a ListStatStore interface. It would have to pass it down to some other functions. Then we could just define a "dummy" list stat store at the very top level.

domino14 (changes requested):

domino14:
I left a comment that we should address. Also -- we should write a migration so we can migrate old list stats to this new store. 

I think this shouldn't be merged in until an implementation for the ListStatStore is ready. Otherwise we are going to lose some list stats. So I can take care of writing that once you make the modification above.

Then once the implementation is written we can merge and deploy. At some point in the near future we would have to write and run a script to migrate old list stats.

PR #56, opened 2020-09-09
Style game buttons, remember the form settings

atomiccupcake:
Also, don't list anonymous users by name in presences, and open links to profile in a new window.


atomiccupcake:
Found a bug. Fixing and I'll reopen this.

PR #60, opened 2020-09-14
Widescreen

atomiccupcake:
Widescreen fixes, tablet fixes, score card display is fancy now, buttons around rack now have tooltips, fixed drag to empty rack, fixed chat overflow on tablet


liwords-ui/src/gameroom/board_panel.tsx
```diff
@@ -118,6 +121,34 @@ export const BoardPanel = React.memo((props: Props) => {
...
+    if (evt.getNickname() === props.username) {
+      return;
+    }
+    let boardMessage = null;
+    switch (evt.getType()) {
+      case GameEvent.Type.PASS:
+        boardMessage = `${evt.getNickname()} passed`;
+        break;
+      case GameEvent.Type.EXCHANGE:
+        boardMessage = `${evt.getNickname()} exchanged`;
```
domino14, line 138: perhaps we could say how many tiles they exchanged here -- I can add it?

pkg/gameplay/sought_game.go
```diff
@@ -9,8 +9,8 @@ import (
 	"github.com/domino14/liwords/pkg/entity"
 )
 
-var errAlreadyOpenReq = errors.New("you already have an open match or seek request")
-var errMatchAlreadyExists = errors.New("the user you are trying to match has already matched you")
+var errAlreadyOpenReq = errors.New("You already have an open match or seek request")
+var errMatchAlreadyExists = errors.New("The user you are trying to match has already matched you")
```
domino14, line 13: This is ok for now. In Go it is standard to start errors with a lowercase letter since they're not intended to be as user visible as we are making them.

In the future what I want to do is return strings like `errors.New("open-match-or-seek-req")`. I have a few other places in Go where I do this. Then on the front end we can have internationalization code that transforms those into full status messages in whatever language.

domino14 (approved):

PR #71, opened 2020-10-01
Fix typing extratime

atomiccupcake:
Make buttons disabled when off turn, allow typing to select tiles to exchange [as described by Conrad in mocks], add key commands for challenge [3], exchange modal [4], exchange all [$], and pass [2]. Hovering over those buttons shows that number. Fix pass/challenge notice at end of game and add clickable links to it. In that state, the pass button is colored as primary.


liwords-ui/src/gameroom/board_panel.tsx
```diff
@@ -340,26 +407,42 @@ export const BoardPanel = React.memo((props: Props) => {
...
+  };
+
+  const handleExchangeModalOk = (exchangedTiles: string) => {
+    setExchangeModalVisible(false);
+    makeMove('exchange', exchangedTiles);
+  };
+
+  const isMyTurn = () => {
+    const iam = gameContext.nickToPlayerOrder[props.username];
+    return iam && iam && iam === `p${gameContext.onturn}`;
```
domino14, line 421: one too many `iam`s?

liwords-ui/src/gameroom/board_panel.tsx
```diff
@@ -125,6 +130,44 @@ export const BoardPanel = React.memo((props: Props) => {
     };
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
+  useEffect(() => {}, [props.board]);
```
domino14, line 133: what's this for?

atomiccupcake: Might be a merge error, might be something I forgot I started. Removing.

PR #76, opened 2020-10-04
relocate tentative score

andy-k:
This pull request moves the tentative score, currently always to the left of the first tentative tile, to just before the first tile of the main word, which can either be an empty square or just outside the board, but never a non-empty square.

For horizontal plays, this resolves the issue when extending an existing word.

For vertical plays, this resolves the issue when hooking to the right of an existing word. (For vertical plays, the tentative score is placed above and not to the left of the first tile.)

When only a single tile is placed, the tentative score is placed according to the placement arrow. Unfortunately if the placement arrow is no longer there, it is ambiguous, and this pull request defaults to horizontal.

Additionally, this pull request makes it so that the tentative score will display even if it is zero (as is the case when placing a blank tile next to another).

### DIFF
--- a/liwords-ui/src/gameroom/board.tsx
+++ b/liwords-ui/src/gameroom/board.tsx
@@ -45,6 +45,7 @@ const Board = React.memo((props: Props) => {
           lastPlayedTiles={props.lastPlayedTiles}
           tentativeTiles={props.tentativeTiles}
           scaleTiles={true}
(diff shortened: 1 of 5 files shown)


domino14:
wow!!

atomiccupcake:
Great work!

atomiccupcake (approved):
I'm going to merge this and then fix that opacity.

PR #78, opened 2020-10-06
fix timeout penalty

andy-k:
It seems that, whenever a game ends in timeout:

The losing player always overtimes by the maximum amount.
Not less, even if no moves in the final minute.
Not more, even if game is abandoned and resumed/adjudicated much later.

And the player on turn is always the losing player.

Additionally, for games allowing less than 5 minutes max overtime (including increment games, which currently have zero max overtime) the penalty should follow the max overtime instead of a hardcoded -50.

This pull request aims to fix cases identified so far relating to the wrong overtime penalty being assessed for games ending in timeout.

~~This pull request does not fix the wrong overtime penalty issue for games ending in other reasons like resigning or losing a triple-challenge.~~

(Update: It fixes the wrong overtime penalty issue for games ending in resigning and triple-challenge, too.)

This pull request does not fix the wrong overtime penalty issue for games ending in any other reasons.

### DIFF
--- a/pkg/gameplay/end.go
+++ b/pkg/gameplay/end.go
@@ -31,20 +31,23 @@ func performEndgameDuties(ctx context.Context, g *entity.Game,
 	evts := []*pb.ServerGameplayEvent{}
 
 	var p0penalty, p1penalty int
-	if g.CachedTimeRemaining(0) < 0 {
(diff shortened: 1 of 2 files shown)


andy-k:
This pull request now fixes the wrong overtime penalty issue for games ending in resigning or triple-challenge, too.

domino14 (approved):
great

PR #80, opened 2020-10-06
do not clear tiles

andy-k:
When you premove and your opponent moves, your tiles get sent back to your rack only if the move affects the premove (by placing a tile where a tentative tile is, or by changing the hooks, which may also happen if the opponent challenges off a word).


atomiccupcake (approved):
This is fantastic. Great job!

PR #81, opened 2020-10-07
do not wrap game controls

andy-k:
After adding the best button ever

Screenshot 2020-10-07 at 14 48 42

we need to disable wrapping

Screenshot 2020-10-07 at 14 48 51


domino14:
Oops, I didn't realize this would break things. We could also just go back to RESIGN as this was kind of added as a joke -- although I do like the connotation of ragequit as resigning shouldn't be encouraged.. I think.

atomiccupcake (approved):
I think this is correct anyway, so I'm merging. Even though that button will be put back to resign almost certainly. :D

PR #90, opened 2020-10-08
Issue 84 - Disable double click of rematch button.

atomiccupcake:
Prevents double clicking rematch but this is nearly impossible to replicate locally, so I kinda wonder if the bot (or different instances of it??) isn't picking up twice in a race condition.

### DIFF
--- a/liwords-ui/src/gameroom/game_controls.tsx
+++ b/liwords-ui/src/gameroom/game_controls.tsx
@@ -1,7 +1,7 @@
 import React, { useState } from 'react';
 import { Button, Popconfirm } from 'antd';
 
-type Props = {
+export type Props = {
   exchangeAllowed?: boolean;
   finalPassOrChallenge?: boolean;
   myTurn?: boolean;
@@ -97,6 +97,7 @@ const EndGameControls = (props: EGCProps) => {
       {props.showRematch && !rematchDisabled && (
         <Button
           type="primary"
+          data-testid="rematch-button"
           onClick={() => {
             setRematchDisabled(true);
             if (!rematchDisabled) {
--- a/liwords-ui/src/gameroom/game_controls.tsx
+++ b/liwords-ui/src/gameroom/game_controls.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import { Button, Popconfirm } from 'antd';
 
 type Props = {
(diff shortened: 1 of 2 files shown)


liwords-ui/src/gameroom/game_controls.tsx
```diff
@@ -87,17 +87,28 @@ type EGCProps = {
...
+      <Button>Options</Button>
+      <Button onClick={props.onExamine}>Export GCG</Button>
+      <Button onClick={() => window.location.replace('/')}>Exit</Button>
+      {props.showRematch && !rematchDisabled && (
+        <Button
+          type="primary"
+          onClick={() => {
+            setRematchDisabled(true);
+            if (!rematchDisabled) {
+              props.onRematch();
```
andy-k, line 103: `setRematchDisabled` isn't immediate, `props.onRematch()` will be called while `rematchDisabled` is still false. and some time later, there would be a rerender with the updated `rematchDisabled`.

i'd suggest to use `useRef`:
```jsx
  const rematchDisabledRef = React.useRef(false);
  return (
    <div className="game-controls">
      <Button>Options</Button>
      <Button onClick={props.onExamine}>Export GCG</Button>
      <Button onClick={() => window.location.replace('/')}>Exit</Button>
      {props.showRematch && !rematchDisabledRef.current && (
        <Button
          type="primary"
          onClick={() => {
            if (!rematchDisabledRef.current) {
              rematchDisabledRef.current = true;
              props.onRematch();
            }
          }}
        >
```

andy-k: (but with `useRef` alone, `EndGameControls` isn't rerendered, and the button remains shown, so maybe use both)

atomiccupcake: I don't think this is correct. React flushes the mouse event queue when state is set. https://github.com/facebook/react/issues/11171#issuecomment-357945371

We could use some front end test though, so I'll write a test to confirm.

andy-k: given that the original code is
```jsx
            setRematchDisabled(true); // 1
            if (!rematchDisabled) { // 2
              props.onRematch(); // 3
```
it's clear that for (3) happens, (2) must have taken place before (1).

andy-k: Interesting, thanks for the reference. Although `setRematchDisabled` will run after the `if`, it seems it's guaranteed to take place before the next click.

So you're right, the concern I brought up is invalid, and this implementation should work. 👍

atomiccupcake: > here, for (3) to happen, (2) must have taken place (and decided that `rematchDisabled` is false) before (1) (sets it to true).
> 
> conversely if (1) happened immediately, (2) would have prevented (3) from running.

Yeah, and the next time a click comes in and the function runs, setState to true is guaranteed by React to have occurred already. But I added a jest test anyway, which will prevent this from being undone accidentally in the future. We need soooo many more component tests.

domino14 (approved):

PR #91, opened 2020-10-08
Logout

atomiccupcake:
Setting this as a draft, because the API endpoint doesn't seem to actually work. I get a 500 error pq: relation "sessions" does not exist. Maybe my database is out of date or something? Feel free to take over the branch. Everything is styled. Moved latency to hovering on the logo instead.

### DIFF
--- a/liwords-ui/src/App.scss
+++ b/liwords-ui/src/App.scss
@@ -1,20 +1,46 @@
 @import 'base';
-.App {
+body {
   background-color: white;
-  /* Prevent squashing board until we get to responsive for mobile */
-  a, button.link {
+}
+
+.ant-dropdown.user-menu {
+  background: white;
+  padding: 6px 24px 6px 12px;
+  min-width: 200px;
+  box-shadow: $card-shadow-lower;
+}
+a, button.link {
+  cursor: pointer;
+  color: $color-primary-dark;
+  font-weight: bold;
+  border: 0;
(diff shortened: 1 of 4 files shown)


domino14:
I merged in andy's fix.  @atomiccupcake could you rebase from master and see if the logout works fine?

domino14 (approved):

atomiccupcake:
> I merged in andy's fix. @atomiccupcake could you rebase from master and see if the logout works fine? Once you test it let me know so I can deploy the previous fix to the backend first.

Not really.

First, I stay logged in until I refresh the browser or get redirected to another . (I can send a chat message as myself or start a game, but I'm logged out when I get TO the game). I can force a refresh or something in the browser.

I am also still seeing those logged out users in presences, long after the refresh. That could be my local don't log people out when the socket drops override. I'll remove that and see if it fixes this part.

domino14:
yeah Logout should definitely force a refresh. Deleting the session doesn't actually disconnect the socket connection which is why you can still chat.

atomiccupcake:
Okay. I refresh the page and that successfully keeps the user from interacting as their self, but something is going to need to be done about presences/redis on your end. It's keeping everybody. If you want to merge this and let that bug be outstanding, it's up to you.

domino14:
Ok great. I'll merge this by tomorrow.

PR #96, opened 2020-10-08
fix backend logout endpoint

andy-k:
the reason #91 failed with 500 is it was trying to `delete from sessions` when it should be `delete from db_sessions`.

### DIFF
--- a/pkg/stores/session/db.go
+++ b/pkg/stores/session/db.go
@@ -98,5 +98,6 @@ func (s *DBStore) Delete(ctx context.Context, sess *entity.Session) error {
 	if sess.ID == "" {
 		return errors.New("session has a blank ID, cannot be deleted")
 	}
-	return s.db.Delete(sess).Error
+	// We want to delete from db_sessions, not delete from sessions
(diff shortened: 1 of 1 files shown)


pkg/stores/session/db.go
```diff
@@ -98,5 +98,6 @@ func (s *DBStore) Delete(ctx context.Context, sess *entity.Session) error {
 	if sess.ID == "" {
 		return errors.New("session has a blank ID, cannot be deleted")
 	}
-	return s.db.Delete(sess).Error
+	// We want to delete from db_sessions, not delete from sessions
+	return s.db.Delete(&dbSession{UUID: sess.ID}).Error
```
domino14, line 102: I really must not have tested this as well as I thought I did. I was pretty sure I tested it with curl/postman, but you're right, this was totally wrong lol.

atomiccupcake (approved):

pkg/stores/session/db.go
atomiccupcake: Yay!

PR #119, opened 2020-10-11
implement drawing on board

andy-k:
For streamers.

Drawings are local.

Screenshot 2020-10-12 at 01 33 27

In this proof-of-concept:
- Right-drag to draw
- Right-click 3x to clear

⚠️ Note: this is not production-quality code, and it's likely someone else more skilled will have to optimize it 😉


liwords-ui/src/gameroom/board.tsx
```diff
@@ -26,13 +26,159 @@ type Props = {
...
+              pic.picture[pic.picture.length - (i + 1)][0].x ===
+                pic.picture[pic.picture.length - 1][0].x &&
+              pic.picture[pic.picture.length - (i + 1)][0].y ===
+                pic.picture[pic.picture.length - 1][0].y
+            )
+          )
+            break;
+        }
+        if (i === howMany) {
+          return { ...pic, drawing: false, picture: [] };
```
domino14, line 131: How does this work? Is there some criteria for how fast the three right-clicks must be?

andy-k: No time limit but they must be at the exact same spot.
Alternatively Shift+RightClick also clears if there's a drawing (otherwise I'm letting it open the context menu)

atomiccupcake: This is neat, but pretty clunky UX to draw with the right mouse pressed. Let's move it to its own component, not call in in yet, and work on it some more when we get time. I'd like it to eventually be something that has to be turned on first (maybe with like a key command), because it will definitely confuse the average user if they right click by accident and can't figure out how to remove.)

atomiccupcake (changes requested):
Please move to a separate component and don't enable yet. Needs work, but I'd like what you have to be in the code base for future work.

liwords-ui/src/gameroom/board.tsx
andy-k: Refactored accordingly. When we've figured out how we want to enable it, it's just a `setEnableDrawing(true);` away.

andy-k:
Recent changes:
- Optimized the code. Performance is better now.
- Added some new features.
- Disabled by default. Use the keyboard to enable.
- Changed the shortcuts so they're harder to trigger accidentally.

Keyboard keys:
- Toggle enabled/disabled: `00`
- Undo: `0U`
- Wipe: `0W` (same as right-click 3x. No more shift+rightclick)
- Change pen color: `0R` `0G` `0B` `0Y`
- Eraser: `0E`

No visual feedback for now other than `console.log`.

atomiccupcake:
This is so cool, but we'd like the main enable/disable to require toggling on with a localStorage flag. ('enableScreenDrawing', or something.) 

The idea with new features and things that change the UX is that the team and our UX designer need a chance to decide how the feature should work. (Feel free to tell anyone that might want to use the feature before then how to put localStorage.setItem('enableScreenDrawing', true); in their console.)

atomiccupcake (approved):

PR #126, opened 2020-10-13
Nicer registration page

atomiccupcake:
image


andy-k:
looks awesome and inviting!

totally optional suggestions:
- [ ] make error on the email field not re-wrap the following fields somehow (image 1)
- [ ] make password field have the same height as the other 3 textboxes, currently it's a few pixels taller (image 1, also when not in error)
- [ ] make error on the password field have the same vertical distance (image 1)
- [ ] make errors have consistent sentence end marks (currently some use `!`, some use `.`, and one uses neither) (image 1)
- [ ] make the checkbox border (both `:hover :checked` and when animating) exclude the `margin-right` (image 2)
- [ ] add `:focus` indicator when tabbing on keyboard to distinguish if pressing space would toggle the checked checkbox or click button (there is `:focus` for unchecked checkbox, but it isn't visible when it's checked) (image 2)
- [ ] add keyboard `tabIndex` access for "toggle password field visibility" feature
- [ ] add "confirm password" field that must match the other box

image 1:
Screenshot 2020-10-13 at 17 03 35

image 2:
Screenshot 2020-10-13 at 17 06 17

atomiccupcake:
> * [ ]  make error on the email field not re-wrap the following fields somehow (image 1)
> * [ ]  make password field have the same height as the other 3 textboxes, currently it's a few pixels taller (image 1, also when not in error)
> * [ ]  make error on the password field have the same vertical distance (image 1)
> * [ ]  make errors have consistent sentence end marks (currently some use `!`, some use `.`, and one uses neither) (image 1)
> * [ ]  make the checkbox border (both `:hover :checked` and when animating) exclude the `margin-right` (image 2)
> * [ ]  add `:focus` indicator when tabbing on keyboard to distinguish if pressing space would toggle the checked checkbox or click button (there is `:focus` for unchecked checkbox, but it isn't visible when it's checked) (image 2)
> * [ ]  add keyboard `tabIndex` access for "toggle password field visibility" feature
> * [ ]  add "confirm password" field that must match the other box

Great finds, Andy! A confirm password field is considered a UX anti-pattern that show is supposed to replace, but Conrad also included one in the mocks, so I'm going to confirm with him before adding one in. Also not sure I can add a tabindex to the show button without redoing the functionality by hand, since Ant-d is doing all of that, but I'll try, otherwise we can add it as a new request in issues.

domino14 (approved):

PR #127, opened 2020-10-13
optimize rendering

andy-k:
an attempt to
- invert the socket implementation into a component that can useContext
- split the StoreContext into several chunks so that components can skip rerendering when only irrelevant chunks change
- separate lag (previously sharing with loginState) to its own context
- add a number of useCallback and memo to reap some benefit


atomiccupcake (approved):
This looks awesome in my testing, but I'll wait for Cesar test and merge, since most of the socket/store code is his.

domino14 (approved):

PR #117, opened 2020-10-11
put player names in gcg filename

andy-k:
if both players have "simple" nicknames, export gcg as `player1-player2-gameId.gcg` instead of `gameId.gcg`.

PR #136, opened 2020-10-14
gcg support more usernames

andy-k:
in #117 usernames were added to gcg files if they meet the validation criteria (since we don't want to create invalid filenames), which was just digits and letters.

in #126 support for `.`, `-`, `_` were added.

this allows newly supported usernames with those characters to also have their usernames in the gcg.

this also means pathological usernames like `.profile` (bad for linux/mac) and `nul.something` (bad for windows) should be prevented separately.

and both "`a-b` vs `c`" and "`a` vs `b-c`" would save as `a-b-c-gameid.gcg`


atomiccupcake (approved):
Nice catch. We've been allowing these usernames from the backers, the registration form is only now reflecting.

PR #140, opened 2020-10-14
Fixes alignment and ratings icon in game history

atomiccupcake:


liwords-ui/src/profile/games_history.tsx
```diff
@@ -49,7 +49,7 @@ export const GamesHistoryCard = React.memo((props: Props) => {
             <span className={`challenge-rule mode_${challenge}`}>
```
domino14, line 49: let's change the challenge rule to use +5, +10 too (instead of -5, -10)

domino14 (changes requested):
small constant fix

domino14 (approved):

PR #146, opened 2020-10-15
implement examine

andy-k:


liwords-ui/src/gameroom/game_controls.tsx
```diff
@@ -13,23 +106,38 @@ export type Props = {
...
 
+  if (props.observer) {
+    return (
+      <div className="game-controls">
+        <Button onClick={props.onExamine}>Examine</Button>
+      </div>
+    );
+  }
+
+  // Temporary dead code.
```
domino14, line 140: should we remove it?

liwords-ui/src/gameroom/game_controls.tsx
```diff
@@ -85,14 +193,16 @@ type EGCProps = {
...
 };
 
 const EndGameControls = (props: EGCProps) => {
   const [rematchDisabled, setRematchDisabled] = useState(false);
   return (
     <div className="game-controls">
       <Button>Options</Button>
-      <Button onClick={props.onExamine}>Export GCG</Button>
+      <Button onClick={props.onExamine}>Examine</Button>
+      <Button onClick={props.onExportGCG}>Export GCG</Button>
```
domino14, line 205: we might want to make this part of Examine, so one of the options there. It's probably ok for right now.

atomiccupcake (approved):
This has a small styling issue (buttons should be more spread out as in mocks) but that's easy to fix later.

PR #155, opened 2020-10-16
do not type to search

andy-k:
"Firefox PC, type-ahead find interferes with typing onto the board"

apparently firefox has this thing in Preferences-General-Browsing, and it's making it painful to type to place tiles

Screenshot 2020-10-16 at 21 18 44

### DIFF
--- a/liwords-ui/src/gameroom/board_panel.tsx
+++ b/liwords-ui/src/gameroom/board_panel.tsx
@@ -764,6 +764,11 @@ export const BoardPanel = React.memo((props: Props) => {
       keydown,
     ]
   );
+  // Just put this in onKeyPress to block all typeable keys so that typos from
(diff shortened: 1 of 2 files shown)


liwords-ui/src/gameroom/board_panel.tsx
```diff
@@ -493,6 +501,7 @@ export const BoardPanel = React.memo((props: Props) => {
       if (handlerReturn === null) {
         return;
       }
+      evt.preventDefault();
```
domino14, line 504: could we just put one blanket `evt.preventDefault()`  around line 435?

andy-k: wouldn't be a good user experience to unnecessarily block PgDn or Tab (for example)

PR #166, opened 2020-10-17
do not reload

andy-k:
make this a single page application instead of reloading.
this way we can make more sounds because it'll be on a page user has interacted in.

added a way to replace the whole store and (indirectly) reconnect the socket.

most page transitions need it. applied it on all for simplicity. (profile page, about page may not actually need it)


atomiccupcake:
Seeing quite a few issues.

Presences not loading in lobby most of the time (the players drop down at top of chat)
Active games list (watch tab) isn’t showing games at all unless they started while the user had the lobby open when they started
Return to game button mode not happening at bottom of games list
Match requests aren’t visible to receiver if they have a second tab or didn't have the lobby tab open when the match was sent (like a rematch comes in and they go to the lobby to see - it's not there)

atomiccupcake (changes requested):
Seeing quite a few issues.

Presences not loading in lobby most of the time (the players drop down at top of chat)
Active games list (watch tab) isn’t showing games at all unless they started while the user had the lobby open when they started
Return to game button mode not happening at bottom of games list
Match requests aren’t visible to receiver if they have a second tab or didn't have the lobby tab open when the match was sent (like a rematch comes in and they go to the lobby to see - it's not there)

domino14:
- I could not reproduce `Active games list (watch tab) isn’t showing games at all...` (started a game against the bot, and other user was in the lobby and it shows up in Watch. When I refresh other user, game still shows up)
   
   I also tried never being present in the lobby until the other user starts the game and it's still ok.

- The presences seem ok from what I can tell. Are you sure you don't have some users blocked locally @atomiccupcake ? I had that issue actually and then once I unblocked a local player it was fine.

- Return to Game at the bottom showed up ok when I go back to the lobby

- I saw the "Return to game with MacondoBot" in the lobby, the big blue button at the bottom, after I had resigned my game against the bot, and gone back to the lobby. However, I can't reproduce it. Maybe I'm wrong, but this might also maybe trigger some thought?

- I also did not see any issues with match requests, and I tried your reproduction steps @atomiccupcake 

The only issue I noticed was that one that I can't reproduce, and I'm not 100% sure that it's there. 

Is it possible your local branch was somehow corrupted?

atomiccupcake (approved):

atomiccupcake:
Yeah, the database had some bad data. I cleared it out and everything seems okay now.

liwords-ui/src/gameroom/pool.tsx
```diff
@@ -103,9 +104,9 @@ const Pool = React.memo((props: Props) => {
       placement="bottomRight"
       overlayClassName="format-dropdown"
     >
-      <a href="/" onClick={(e) => e.preventDefault()}>
+      <Link to="/" onClick={(e) => e.preventDefault()}>
         Rearrange
-      </a>
+      </Link>
```
domino14, line 109: hmmm.. will this do the right thing? I wonder if this is the right way to have a link there...

liwords-ui/src/store/store.tsx
```diff
@@ -756,6 +763,20 @@ export const Store = ({ children, ...props }: Props) => {
...
+const ResetStoreContext = createContext({ resetStore: defaultFunction });
+export const useResetStoreContext = () => useContext(ResetStoreContext);
+
+export const Store = ({ children }: { children: React.ReactNode }) => {
+  // In JS the | 0 loops within int32 and avoids reaching Number.MAX_SAFE_INTEGER.
+  const [storeId, resetStore] = React.useReducer((n) => (n + 1) | 0, 0);
+
+  return (
+    <ResetStoreContext.Provider value={{ resetStore }}>
+      <RealStore key={storeId} children={children} />
```
domino14, line 775: i really didn't know this could be done -- it essentially just creates an entire new store?

domino14:
awesome

PR #183, opened 2020-10-19
improve exchange

andy-k:
select one tile at a time, deselect all if all instances selected. toggle feature.


domino14:
we'd need to discuss this internally i think as it is a UX design fix partially

atomiccupcake (commented):
Awesome. Conrad says go ahead and do this and see how it goes. Thanks!

PR #184, opened 2020-10-19
misc cleanup

andy-k:


liwords-ui/src/gameroom/table.tsx
```diff
@@ -138,7 +138,7 @@ export const Table = React.memo((props: Props) => {
 
   useEffect(() => {
     if (gameContext.playState === PlayState.GAME_OVER || isObserver) {
-      return () => {};
+      return;
```
domino14, line 141: my linter complained about not returning a function (consistent returns). Do we have the same linter settings? They should be checked into this repo.

liwords-ui/src/gameroom/table.tsx
```diff
@@ -209,9 +201,7 @@ export const Table = React.memo((props: Props) => {
           }
         )
         .then((streakresp) => {
-          if (stillMountedRef.current) {
-            setStreakGameInfo(streakresp.data.game_info);
-          }
+          setStreakGameInfo(streakresp.data.game_info);
```
domino14, line 204: this one would probably need to be protected as it's an XHR - what do you think?

andy-k: it is. we're no longer using `React.useState` directly

domino14 (commented):
as you mentioned, i am most worried about the anti-pattern of checking whether stuff is mounted. What is the alternative -- to cancel all AJAX requests in the right places?

domino14:
It still seems worth merging as it's a clean way of setting state, and it seems to work, so i'll merge it.

PR #192, opened 2020-10-21
render blank l differently

andy-k:
changing blank `l`

Screenshot 2020-10-22 at 04 19 11

to

Screenshot 2020-10-22 at 04 14 22


domino14:
i think @atomiccupcake's design for blanks will be implemented

PR #193, opened 2020-10-22
Branded blank

atomiccupcake:
Undesignated still are completely blank.

image

PR #200, opened 2020-10-23
center the blank letter

andy-k:
the one in #193 wasn't perfectly centered


domino14:
Screen Shot 2020-10-22 at 11 47 23 PM

it seems like it's a little too far down vertically. 0 degree rotation above to make it a little easier to see.

Screen Shot 2020-10-22 at 11 47 54 PM

domino14:
I think it's because the other letters are a little high (to make room for the score)

domino14:
actually i think once it's angled it's fine..

PR #201, opened 2020-10-23
support drag to gap between tiles on rack

andy-k:
the tiles on the rack has a margin, previously dragging to that margin wasn't working.

https://github.com/domino14/liwords/blob/eab06f5eaa6d7ef1ef69a2c6061242796958490e/liwords-ui/src/gameroom/scss/gameroom.scss#L413-L416


domino14 (approved):

domino14:
this works great; shuffling seems smoother

PR #215, opened 2020-10-28
lobby chat in game

andy-k:
- includes a fix for an existing bug exposed by this implementation
- not the ideal implementation for this feature

Screenshot 2020-10-28 at 15 17 30


atomiccupcake (changes requested):
https://www.figma.com/file/fjhPTM7M4tVPhde8xTDu82/Woog-for-Engineering?node-id=73%3A1099 This is the design for changing between chats. Sorry for the lack of clarity. I shall try to explain it.

At the top of the chat window there should be a link with a < that says "All chats". Clicking it will go to this view https://www.figma.com/file/fjhPTM7M4tVPhde8xTDu82/Woog-for-Engineering?node-id=73%3A37 

Lobby chat would be a choice under YOUR CHATS (the only one currently, until we have direct messages -- and until then there's no "+ New chat" button). It would have an L avatar. The last chat that came in is previewed in blue if it came in since you switched your view to that channel.

Clicking on a choice goes back to the main view. 

Feel free to put a blue dot to the right of  "All chats" if something comes in on the unselected chats, but only if it's coming from the main important chat (game/observer) while you're in the lobby/etc. That isn't in the mocks, and Conrad may change this part.

andy-k:
thanks for the review. it seems the backend isn't ready to support this feature, so shelving it for now.

PR #233, opened 2020-11-01
move macondo to Worker

andy-k:
caveat: this relies on a new .js in public/.
(seems it cannot be put in src/ because of Reasons.)

this .js:
- may need a special caching/hashing policy.
- is, by definition, not type-safe.
- is not preprocessed/transpiled/minified.

### DIFF
--- a/liwords-ui/public/wasm/macondo.js
+++ b/liwords-ui/public/wasm/macondo.js
@@ -1,0 +1,101 @@
+// This script must be pure JS outside React/webpack.
+// So, it is open-source.
+
+(() => {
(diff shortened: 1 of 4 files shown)


liwords-ui/public/wasm/macondo.js
```diff
@@ -0,0 +1,101 @@
+// This script must be pure JS outside React/webpack.
+// So, it is open-source.
```
domino14, line 2: we're open-source anyway :) you mean it's not minified/etc?

domino14:
wow - thank you for looking into this. This will help a lot with simming in the future.

PR #241, opened 2020-11-09
improve analyzer ux

andy-k:
improves #240 with
- some code quality improvements
- avoid placing exchanged tiles on existing tiles (tiles from rack vanished if there's a tile at 1A)
- prevent showing analysis for a different turn (race condition)

removing the extra zero moves relies on macondo not returning them: https://github.com/domino14/macondo/pull/109

### DIFF
--- a/liwords-ui/src/gameroom/analyzer.tsx
+++ b/liwords-ui/src/gameroom/analyzer.tsx
@@ -5,7 +5,6 @@ import {
   useExaminableGameContextStoreContext,
   useTentativeTileContext,
 } from '../store/store';
-import { Unrace } from '../utils/unrace';
 import { getMacondo } from '../wasm/loader';
 import { useMountedState } from '../utils/mounted';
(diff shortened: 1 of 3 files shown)


liwords-ui/src/gameroom/analyzer.tsx
```diff
@@ -142,39 +115,70 @@ export const Analyzer = React.memo((props: AnalyzerProps) => {
...
+  const examinerId = useRef(0);
+
   useEffect(() => {
-    if (moves.length > 0) {
-      setExaminerLoading(false);
-    }
-  }, [moves, setMoves, examinerLoading, setExaminerLoading]);
+    setExaminerLoading(false);
+    examinerId.current = (examinerId.current + 1) | 0;
+  }, [moves]);
```
domino14, line 123: these extra dependencies are not required because they're `setState`s right?

Issue #114, opened 2020-10-11
Firefox/Safari don't make sounds sometimes


domino14:

domino14:
it seems Firefox has strict rules governing interactions with the DOM. it makes sounds when you are playing moves but not when you first enter a table, as it is a new URL and it has refreshed...

PR #248, opened 2020-11-12
sound workaround for sound

andy-k:
this might mitigate #114 

sounds that are requested to play before user interacts with the tab, will play when the user does.


liwords-ui/src/sound/boop.ts
```diff
@@ -1,42 +1,83 @@
...
-  endgameSound,
-  woofSound,
-};
+import { Unrace } from '../utils/unrace';
 
-const playSound = (soundName: string) => {
-  const audio = playableSounds[soundName];
-  if (!audio) {
-    throw new TypeError(`unsupported sound: ${soundName}`);
+class Booper {
```
domino14, line 3: hehe

PR #263, opened 2020-11-14
do not auto-scroll chat

andy-k:
adding a subtle indicator for unread chat. (discord doesn't need it because it always shows scrollbar.)
auto-scrolling resumes when user scrolls to end.

Screenshot 2020-11-14 at 22 26 27

### DIFF
--- a/liwords-ui/src/chat/chat.tsx
+++ b/liwords-ui/src/chat/chat.tsx
@@ -187,7 +187,7 @@ export const Chat = React.memo((props: Props) => {
           <div className={`chat-context${hasScroll ? ' scrolling' : ''}`}>
             <p>
               {props.description}
-              {hasUnreadChat && '*'}
+              {hasUnreadChat && ' •'}
             </p>
             {presenceCount ? (
               <>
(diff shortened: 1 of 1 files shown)


liwords-ui/src/chat/chat.tsx
```diff
@@ -117,7 +185,10 @@ export const Chat = React.memo((props: Props) => {
         </TabPane>
         <TabPane tab="Chat" key="CHAT">
           <div className={`chat-context${hasScroll ? ' scrolling' : ''}`}>
-            <p>{props.description}</p>
+            <p>
+              {props.description}
+              {hasUnreadChat && '*'}
+            </p>
```
atomiccupcake, line 191: Could we use • instead of * and put a space before it, so it's more pleasant looking?

atomiccupcake (changes requested):

liwords-ui/src/chat/chat.tsx
atomiccupcake: Additionally, the math seems to not work for tablet view and maybe phone. When I scroll to the bottom after a message comes in, the indicator doesn't vanish

atomiccupcake: Bug looks like a testing error with windows not zoomed to 100%, so way too edge case to fix. Approving.

atomiccupcake (approved):

Issue #210, opened 2020-10-25
need more retries on failure


domino14:
Had a temporary bad connection; I joined a game and presumably the socket connection was not acquired, or the get metadata somehow timed out. So the game just sat there and would probably have been automatically aborted if i hadn't refreshed.

domino14:
slightly worried that it was a 500 on the backend, but i didn't see that in the ELB, and also connecting to other sites was giving me timeouts (like AWS and Facebook).

domino14:
This happens in the wild relatively often. I tracked down one particular event:
- WK matched Saraberry
- Saraberry accepted
- Saraberry's client sent the realm request 
- WK's client never sent the realm request

The socket logs showed that WK's socket disconnected at the right time, but never reconnected. So the redirect presumably worked (disconnecting the original connection) but the new connection took at least 40 seconds to come in (he might have refreshed if it never came in). Maybe his browser somehow froze. The ALB logs don't seem to show the new connection timing out or coming it at all until significantly later.

domino14:
socket stuff seems to be ok for now.

PR #272, opened 2020-11-18
fix socket timing

andy-k:
- clean up socket eslint. this uncovered that the socket url was dependent on location.pathname.
- resetStore implicitly when (and after) path changes. currently it may be done before.
- remove most explicit resetStore(), keeping only the cases where there is no path change (logout and "login on /").
- add encodeURIComponent. just a tech debt.
- remove redirGame to simplify code flow. updating the url is already doing the right thing.

### DIFF
--- a/liwords-ui/src/App.tsx
+++ b/liwords-ui/src/App.tsx
@@ -1,5 +1,5 @@
 import React, { useCallback, useEffect, useRef } from 'react';
-import { Route, Switch, useHistory, useLocation } from 'react-router-dom';
+import { Route, Switch, useLocation } from 'react-router-dom';
 import { useMountedState } from './utils/mounted';
 import './App.scss';
 import axios from 'axios';
@@ -10,7 +10,6 @@ import TileImages from './gameroom/tile_images';
 import { Lobby } from './lobby/lobby';
 import {
   useExcludedPlayersStoreContext,
-  useRedirGameStoreContext,
   useResetStoreContext,
 } from './store/store';
 
@@ -35,7 +34,6 @@ const App = React.memo(() => {
   const { useState } = useMountedState();
 
   const { setExcludedPlayers } = useExcludedPlayersStoreContext();
-  const { redirGame, setRedirGame } = useRedirGameStoreContext();
   const { resetStore } = useResetStoreContext();
   const [shouldDisconnect, setShouldDisconnect] = useState(false);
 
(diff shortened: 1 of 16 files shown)


andy-k:
also attempt to mitigate #210 by setting reconnect options (it doesn't fully solve it, still need client-initiated ping)

liwords-ui/src/App.tsx
```diff
@@ -49,10 +49,18 @@ const App = React.memo(() => {
...
+  }, [history, redirGame, setRedirGame]);
+
+  const location = useLocation();
+  const knownLocation = useRef(location.pathname); // Remember the location on first render.
+  useEffect(() => {
+    if (knownLocation.current !== location.pathname) {
+      resetStore();
     }
-  }, [history, redirGame, resetStore, setRedirGame]);
+  }, [location.pathname, resetStore]);
```
domino14, line 52: I don't totally understand how useRef works here. Does `knownLocation` ever change after the first render? What if we start at `/` then go to `/game/foo` then go back to `/`? Wouldn't the `if` statement then fail, and the store won't reset?

liwords-ui/src/store/socket_handlers.ts
```diff
@@ -454,7 +456,7 @@ export const useOnSocketMsg = () => {
               payload: '',
             });
             const gid = nge.getGameId();
-            setRedirGame(gid);
+            historyRef.current.replace(`/game/${encodeURIComponent(gid)}`);
```
domino14, line 459: could you explain why this instead of `history.replace`? I am a little confused about what the refs do here

liwords-ui/src/App.tsx
andy-k: - if we start at `/`, the `knownLocation` for this instance of `<App />` is set to `/`.
- `knownLocation` never changes after the first render.
- the moment we go to `/game/foo`, this instance will `resetStore()`.
- `resetStore()` will unmount this `<App />` and mount a totally new, independent instance of `<App />`.
- when that new instance is first rendered, its `knownLocation` is set to `/game/foo`.
- the old instance will be garbage-collected.
- when we go back to `/`, it is the new instance that will `resetStore()`.

liwords-ui/src/store/socket_handlers.ts
andy-k: using ref avoids having to make `history` a dependency of this `useCallback`. here `historyRef.current` will refer to the latest observed instance of `history` because we reassign it every render.

liwords-ui/src/App.tsx
domino14: got it :+1:

Issue #87, opened 2020-10-08
MVP Match behavior enhancements


domino14:
- [x] Match during a game should show a small pop-up / notification somewhere, but they still have to go to the Lobby to accept it or decline it.
- [x] If both players request a rematch, the second one should just immediately accept the first rematch in the backend.
- [x] Redirect any observers of rematches to the new games as well

PR #284, opened 2020-11-25
click rematch to accept rematch

andy-k:
#283

### DIFF
--- a/liwords-ui/src/gameroom/table.tsx
+++ b/liwords-ui/src/gameroom/table.tsx
@@ -555,6 +555,9 @@ export const Table = React.memo((props: Props) => {
             playerMeta={gameInfo.players}
             tournamentID={gameInfo.tournament_id}
             lexicon={gameInfo.lexicon}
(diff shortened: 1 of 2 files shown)


liwords-ui/src/gameroom/table.tsx
```diff
@@ -555,6 +555,9 @@ export const Table = React.memo((props: Props) => {
             playerMeta={gameInfo.players}
             tournamentID={gameInfo.tournament_id}
             lexicon={gameInfo.lexicon}
+            handleAcceptRematch={
+              rematchRequest.getRematchFor() !== '' ? handleAcceptRematch : null
```
domino14, line 559: i can see an edge case here where you got a rematch for a separate game in a different tab (although, i may already handle that properly with the connection ID stuff). What if we say `getRematchFor() === gameID`

domino14 (approved):

domino14:
close #283 
close #87

PR #287, opened 2020-11-28
P2p messages

domino14:


liwords-ui/src/shared/usernameWithContext.tsx
```diff
@@ -27,6 +37,11 @@ export const UsernameWithContext = (props: UsernameWithContextProps) => {
       {props.userID ? (
         <TheBlocker className="link plain" target={props.userID} tagName="li" />
       ) : null}
+      {!props.omitSendMessage && props.userID ? (
+        <li onClick={() => sendMessage(props.userID!, props.username)}>
```
atomiccupcake, line 41: This all seems to be working well. Let's add a style display = none on this before you merge, since the UX is fairly confusing right now (can't see who you sent the message, to, etc.) And I'll get Conrad's UX built in the next couple of days

atomiccupcake (approved):
I'd hide the button for now, but otherwise this looks good.

liwords-ui/src/shared/usernameWithContext.tsx
domino14: Ok great. I think before merging though, I need to fix a bunch of presence stuff in `liwords-socket`. We can collaborate on this branch in the meantime; feel free to commit to it.. I'll let you know if anything changes.

PR #293, opened 2020-12-02
Tournament Messages

jvc56:


pkg/tournament/service.go
```diff
@@ -30,40 +31,60 @@ func (ts *TournamentService) SetEventChannel(c chan *entity.EventWrapper) {
 }
 
 func (ts *TournamentService) AddDivision(ctx context.Context, req *pb.TournamentDivisionRequest) (*pb.TournamentResponse, error) {
-	err := AddDivision(ctx, ts.tournamentStore, req.Id, req.Division)
+	err := authenticateDirector(ctx, ts, req.Id, false)
+	if err != nil {
+		return nil, err
```
domino14, line 36: please wrap any errors in this service like the others (`twirp.NewError(twirp.Unauthenticated, ...` ) .. otherwise it will result in 500s.

api/proto/realtime/realtime.proto
```diff
@@ -84,8 +85,10 @@ enum MessageType {
   SERVER_MESSAGE = 24;
   READY_FOR_GAME = 25;
   LAG_MEASUREMENT = 26;
-  TOURNAMENT_GAME_ENDED_EVENT = 27;
+  TOURNAMENT_MESSAGE = 27;
```
domino14, line 88: don't reuse this message - can you make it a new one? we can obsolete the old one later (mark it obsolete). reusing messages in pb can really result in issues.

pkg/tournament/service.go
jvc56: Errors wrapped.

api/proto/realtime/realtime.proto
jvc56: Unreused.

pkg/tournament/service.go
domino14: so `twirp.InternalErrorWith` still results in 500s. It is fine to do this, but only if it's an actual internal server error. For the errors you used, are these errors on the server side, or are some of them due to the user doing something wrong? For the ones that are the latter, you'll need to add an error code (like we did with `twirp.Unauthenticated`) that will give you a 400 (for example, twirp.Malformed) see: https://twitchtv.github.io/twirp/docs/errors.html

jvc56: It could be either, should I just make tournament errors return InvalidArgument errors?

domino14: it depends on the error -- if it's actually an error like our database didn't connect or something then it should return a 500. Otherwise it should return InvalidArgument. you should see what the operations look like to determine what kind of error it was, I think.

PR #300, opened 2020-12-06
Mutual Abort

jvc56:
- [x] test old auto-abort behavior (cancellation)
- [x] look into adding game deletion from lobby for tournaments (why was it difficult?)

Cases to consider:
- [x] disallow abort/nudge if there are less than X minutes remaining on person's clock
- [ ] countdown timer for abort/nudge should probably start at a number that is less than the time remaining on the clock 
- [x] likely need to send outstanding abort/nudge events on reconnect. This is to deal with a case where someone has a bad connection, they disconnect, their opponent requests an abort/nudge, and then the player returns and has no idea such an event is outstanding
- [x] need to auto-cancel abort/nudge by making a move on the board. this can probably be implemented with a useEffect on the turn number automatically canceling outstanding events.
- [x] disable abort after some number of turns
- [x] a person whose clock is running shouldn't be nudging anyone
- [x] maybe show when people leave the room
- [x] disallow in bot game
- [x] fix styling to match Figma design
- [x] disallow or modify for tournaments/clubs
- [x] disallow multiple clicks on abort/resign/etc bringing up multiple modals
- [x] abort/nudge sound


pkg/bus/meseeks.go
```diff
@@ -75,6 +75,45 @@ func (b *Bus) seekRequest(ctx context.Context, auth, userID, connID string,
...
+}
+
+func (b *Bus) abortAccepted(ctx context.Context, evt *pb.AbortRequest,
+	userID, connID string) error {
+	req := &pb.AbortRequest{}
+	err := proto.Unmarshal(data, req)
+	if err != nil {
+		return err
+	}
+	return gameplay.AbortGame(ctx, b.gameStore, req.GameId)
```
domino14, line 114: you should change this function to take in one more parameter (game end reason, which is either `CANCELLED` or `ABORTED`). Right now it only does `CANCELLED` -- it gets called when a game never starts at all. Let's add `ABORTED` for this particular case so it differentiates between the two cases. The rest of the code looks pretty good.

domino14:
Design:
Screen Shot 2021-03-13 at 11 36 22 AM

pkg/gameplay/meta_events.go
```diff
@@ -176,9 +177,11 @@ func HandleMetaEvent(ctx context.Context, evt *pb.GameMetaEvent, eventChan chan<
 			}
 		}
 
+		// XXX: Receiver may not be the one on turn, since either player may request abort.
 		onTurn := g.Game.PlayerOnTurn()
 		timeRemaining := g.TimeRemaining(onTurn)
 		log.Debug().Int("timeRemaining", timeRemaining).Int("onturn", onTurn).Msg("timeremaining")
+		// XXX: Should time remaining include overtime?
```
domino14, line 188: this is probably ok for the time being

pkg/gameplay/meta_events.go
```diff
@@ -188,6 +191,7 @@ func HandleMetaEvent(ctx context.Context, evt *pb.GameMetaEvent, eventChan chan<
 			return ErrNotAllowed
 		}
 
+		// XXX: Adjust reasonably based on receiver's remaining time.
```
domino14, line 198: it won't really matter for the time being; game would end before nudge/abort times out so i think it's fine.

pkg/gameplay/meta_events.go
```diff
@@ -167,6 +167,7 @@ func HandleMetaEvent(ctx context.Context, evt *pb.GameMetaEvent, eventChan chan<
 		if entity.LastOutstandingMetaRequest(g.MetaEvents.Events, evt.PlayerId) != nil {
 			return ErrAlreadyOutstandingRequest
 		}
+		// XXX: What if player A sends a meta request and player B sends a different meta request?
```
domino14: handled

PR #301, opened 2020-12-07
Private messaging

atomiccupcake:
Adds private messaging and the ability to use tournament chat while in game. Adds a block list to profile, makes block list refresh on block action. Also cleans up appearance of the new club/tournament page.


liwords-ui/src/chat/chat.tsx
```diff
@@ -50,43 +85,142 @@ export const Chat = React.memo((props: Props) => {
...
+                lastUpdate: parseInt(ch.last_update, 10),
+                hasUpdate: ch.has_update,
+                lastMessage: ch.last_message || '',
+                name: ch.name,
+              };
+            }) || [],
+        };
+        setChatChannels(newChannels);
+      });
+  }, [setChatChannels]);
```
domino14, line 123: i believe that `setX` type state manipulation functions do not need to be specified as dependencies (they never change) and the linter picks up on this. I'll test it locally.

liwords-ui/src/chat/chat.tsx
```diff
@@ -50,43 +85,142 @@ export const Chat = React.memo((props: Props) => {
...
+        {
+          number: 20,
+          offset: 0,
+        },
+        { withCredentials: true }
+      )
+      .then((res) => {
+        const newChannels: ActiveChatChannels.AsObject = {
+          channelsList:
+            res.data.channels.map((ch: any) => {
```
domino14, line 111: We should try to avoid doing the `any`. I know why you are -- it's this protobuf to object stuff and the extremely annoying fact that it sends `display_name` when you invoke it with JSON but the `AsObject` defines it as `displayName`. Maybe there's no real way around this yet. What I've been doing is just defining shadow types that are very similar but with the other type of casing. I have to figure out a better way to do this.

domino14 (approved):

liwords-ui/src/chat/chat.tsx
atomiccupcake: I'll do that

liwords-ui/src/chat/chat.tsx
atomiccupcake: Yeah, it seems pretty random as to when the linter yells at me or doesn't. I'll double check trying to remove these one by one.

atomiccupcake: Nah. It wouldn't let me remove this one.

andy-k: if it's from a `useState` in the same function, it can be removed from dependency. since it's not, it needs to be specified as dependency.
https://github.com/domino14/liwords/blob/ea3587fe00d157a7246687c1b96a2ea7c1d33e39/liwords-ui/src/chat/chat.tsx#L62-L63

PR #306, opened 2020-12-14
progress on multiple api server support

domino14:
Support multiple API servers. This is important during deploy.

- Invalidate game cache if turn sequence numbers don't match. Reloading game from db, even one that is almost done, only takes on the order of 1 millisecond
- Add a `soughtgame` db store so we don't lose seeks during deploys


domino14:
this is going to be handled differently, as in #889 

see #905 eventually

PR #323, opened 2020-12-27
Initial Fontes

jvc56:


pkg/pair/pair.go
```diff
@@ -233,7 +273,7 @@ func weighSwiss(members *entity.UnpairedPoolMembers, i int, j int) int64 {
 
 	// Subtract the spread difference for swiss, since we would like to pair players
 	// that have similar records but large differences in spread.
-	spreadDiffWeight := -int64(utilities.Abs(p1.Spread-p2.Spread))
+	spreadDiffWeight := -int64(utilities.Abs(p1.Spread - p2.Spread))
```
domino14, line 276: `gofmt` ?

jvc56: Must be, I don't remember changing that.

PR #349, opened 2021-01-05
Remove Player

jvc56:


pkg/tournament/classic_division.go
```diff
@@ -500,6 +531,57 @@ func (t *ClassicDivision) AddPlayers(persons *entity.TournamentPersons) error {
...
+		if i == t.CurrentRound {
+			presentResults, err := t.ResultsArePresent(i)
+			if err != nil {
+				return err
+			}
+			resultsArePresent = presentResults
+		}
+		if i > t.CurrentRound || (i == t.CurrentRound && !resultsArePresent) {
+			pm := t.RoundControls[i].PairingMethod
+			if i == t.CurrentRound || pair.IsStandingsIndependent(pm) && pm != entity.Manual {
```
andy-k, line 573: which of these is intended? (by default && is ()'ed before ||)
```suggestion
			if i == t.CurrentRound || (pair.IsStandingsIndependent(pm) && pm != entity.Manual) {
```
or
```suggestion
			if (i == t.CurrentRound || pair.IsStandingsIndependent(pm)) && pm != entity.Manual {
```

jvc56: Ah, thanks for pointing that out. The first of your suggestions is the intended expression. I prefer this suggestion over relying on boolean precedence since it makes the code more clear and I don't have to think about boolean precedence. Thanks for catching that.

PR #355, opened 2021-01-06
start simultaneous games in lobby

andy-k:
- doesn't affect tournament for now, only main lobby
- if player has at least one active game, Play tab lists player's active games, so player can select which one to resume
- when one game is active, player has to click on a text like "Play more games simultaneously (timers will run concurrently)" to be able to start the second concurrent game (they need to click on the text, not the background)
- when two or more games are active, player no longer sees "Resume your game with (whomever)", they'll select from the Resume list instead
- (we'll need to support opening games in new tabs) (it's possible to misclick if the Available Games change)

Screenshot 2021-01-06 at 12 06 58

Screenshot 2021-01-06 at 12 07 09

Screenshot 2021-01-06 at 12 07 21


andy-k:
interim update: reordered the sections so Resume is always first and the other sections will only show until the player agrees to simul.

Screenshot 2021-01-23 at 09 11 32

andy-k:
using ignore link

Screenshot 2021-01-23 at 11 24 21

atomiccupcake:
Please remove the resume section. Thanks.

PR #372, opened 2021-01-11
Display pairings for paired-mode tournaments, letting user select div…

domino14:
…ision and page through rounds. If the user is registered, it will default to their division, and their current round.

### DIFF
--- a/liwords-ui/src/tournament/pairings.tsx
+++ b/liwords-ui/src/tournament/pairings.tsx
@@ -1,0 +1,126 @@
+import React, { ReactNode } from 'react';
+import { useTournamentStoreContext } from '../store/store';
+import { Table, Tag } from 'antd';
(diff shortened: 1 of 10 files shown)


liwords-ui/src/tournament/pairings.tsx
```diff
@@ -0,0 +1,126 @@
+import React, { ReactNode } from 'react';
+import { useTournamentStoreContext } from '../store/store';
+import { Table, Tag } from 'antd';
+import { Division, SinglePairing } from '../store/reducers/tournament_reducer';
+
+const usernameFromPlayerEntry = (p: string) =>
+  p.split(':').length > 0 ? p.split(':')[1] : 'Unknown player';
```
domino14, line 7: not a big deal but should avoid splitting twice. I'll merge anyway, we can handle it after.

PR #375, opened 2021-01-13
settings modal

andy-k:
includes #374 

Screenshot 2021-01-13 at 10 31 04

(feel free to take over and restyle as needed)


domino14:
a new settings modal/options are being designed so I think let's wait for that

PR #406, opened 2021-01-29
Tournament Upgrades

### LINKED ISSUE #387: empty map turns into nil when loaded from the db in tournament division
- create division
- create tournament controls
- restart server
- add players --> panic

### LINKED ISSUE #472: when a new round starts, if you click Watch sometimes it goes to the previous game for one of the involved players
https://www.twitch.tv/videos/956015131?t=00h12m23s

### LINKED ISSUE #477: WATCH button disappeared (in tournament mode)
Every game should have a WATCH button next to it, but they disappeared at some point. Refreshing brings them back.

### LINKED ISSUE #479: I'm Ready button showed up in another tab while i was playing the game in the first tab
i was too afraid to click the button, but it hopefully wouldn't have done anything?

### LINKED ISSUE #481: Removing a player from the tournament does not update the pairings
I had to remove a player (A) after 8 rounds. This was all prior to opening round 9. However, the round 9 pairings still showed the player's previous pairing (on refresh, as well).

Once I added a player to replace them, however, it correctly showed A as forfeiting round 9.

### LINKED ISSUE #569: add more fields to tournament metadata
child mode
disclaimer (caption)
logo 
tile style
board style
hardcoded settings
password

jvc56:
- [x] Make sure clubs still work
- [x] Need to migrate old tournaments
- [x] Figure out how to expire "registrants" table when tournaments end
- [x] Figure out how to flush cache (we might need to have a brief downtime)
- [ ] Disallow players from being in multiple divisions (@jvc56 do this at some point)

fix #481 
fix #479 (probably)
fix #477 (most likely)
fix #472 (most likely)
fix #387 

close #569


pkg/tournament/tournament_classic_test.go
```diff
@@ -264,24 +270,24 @@ func TestClassicDivisionKingOfTheHill(t *testing.T) {
...
 	// Tournament should not be over
 
-	player1 := playerStrings[0]
-	player2 := playerStrings[1]
-	player3 := playerStrings[3]
-	player4 := playerStrings[2]
+	player1 := defaultPlayers.Persons[0].Id
+	player2 := defaultPlayers.Persons[1].Id
+	player3 := defaultPlayers.Persons[3].Id
+	player4 := defaultPlayers.Persons[2].Id
```
andy-k, line 290: (are player3, player4 being [3], [2] instead of [2], [3] intended?)

jvc56: This was a copy paste from another test where order used to matter but doesn't anymore due to changes in the way players are ranked initially. Which is player3 or player4 doesn't matter for this specific case, but it does look wrong and confusing, so I will switch it around to [2], [3] in the next push. Good catch.

scripts/migrations/tournaments/main.go
```diff
@@ -0,0 +1,516 @@
...
+	ids, err := tournamentStore.ListAllIDs(ctx)
+	if err != nil {
+		panic(err)
+	}
+	log.Info().Interface("ids", ids).Msg("listed-game-ids")
+
+	for _, tid := range ids {
+		_, err := oldDatabaseObjectToEntity(ctx, tournamentStore, tid)
+		if err != nil {
+			panic(err)
```
domino14, line 506: is it possible to wrap this entire thing in a transaction? a panic in the middle is scary

jvc56: Yes.

scripts/migrations/tournaments/main.go
```diff
@@ -500,17 +454,91 @@ func main() {
 	}
 	log.Info().Interface("ids", ids).Msg("listed-game-ids")
 
-	for _, tid := range ids {
-		_, err := oldDatabaseObjectToEntity(ctx, tournamentStore, tid)
-		if err != nil {
-			panic(err)
-		}
+	tournamentStore.db.Transaction(func(tx *gorm.DB) error {
```
domino14, line 457: `tournamentStore.db` is a private variable -- how's this compile?

jvc56: I've redefined NewDBStore in this script so I could have access to that field. See above.

liwords-ui/src/api/api.ts
```diff
@@ -12,16 +12,33 @@ interface PBMsg {
...
+
+export const twirpErrToMsg = (err: TwirpError) => {
+  // Twirp always returns JSON error messages no matter what. But since the
+  // responseType is set to `arraybuffer` above it is annoying to deal with.
+  // This function turns it into the JSON-encoded string that it is and
+  // extracts the error message.
+  const errJSON = String.fromCharCode.apply(
+    null,
+    Array.from(new Uint8Array(err.response.data))
+  );
```
andy-k, line 42: to handle multibyte utf8 codepoints:
```suggestion
  const errJSON = new TextDecoder().decode(err.response.data);
```

pkg/tournament/tournament.go
```diff
@@ -550,30 +788,95 @@ func SetResult(ctx context.Context,
...
+		if !isReady {
+			return fmt.Errorf("cannot start round %d for division %s because the round is not ready", round, division)
+		}
 	}
-	if !t.IsStarted {
-		return fmt.Errorf("cannot start tournament %s with no divisions", t.Name)
+
+	for division := range t.Divisions {
+		t.IsStarted = true
+		err := StartRoundCountdown(ctx, ts, id, division, round, false, true)
```
andy-k, line 875: just noting here
- `StartAllRoundCountdowns` calls `ts.Get` but does not lock
https://github.com/domino14/liwords/blob/5af38e6d1aa446e15bdeb1a2aca40e0ce981add8/pkg/tournament/tournament.go#L842-L848
- `StartAllRoundCooldowns` calls `StartRoundCountdown` (here) for each division
- `StartRoundCooldown` calls `ts.Get` again and locks/unlocks each time
https://github.com/domino14/liwords/blob/5af38e6d1aa446e15bdeb1a2aca40e0ce981add8/pkg/tournament/tournament.go#L893-L901
- but `ts.Get` is not obligated to return the same object
- it's not immediately clear that this `StartAllRoundCooldowns` depends on `ts` implementing a caching mechanism

@jvc56

PR #409, opened 2021-01-31
Raise Minimum Length of Username to 3 Characters

jvc56:


pkg/registration/register.go
```diff
@@ -15,8 +15,8 @@ import (
 func RegisterUser(ctx context.Context, username string, password string, email string,
 	userStore user.Store, bot bool, argonConfig config.ArgonConfig) error {
 	// username = strings.Rep
-	if len(username) < 1 || len(username) > 20 {
-		return errors.New("username must be between 1 and 20 letters in length")
+	if len(username) < 3 || len(username) > 20 {
+		return errors.New("username must be between 3 and 20 letters in length")
```
andy-k, line 19: this would show only after submission, like this:

Screenshot 2021-02-01 at 08 00 06

consider also adjusting [`register.tsx`](https://github.com/domino14/liwords/blob/a1bebf117333735e53d1abe9de15a2add8206d68/liwords-ui/src/lobby/register.tsx#L98-L117) to get as-they-type feedback:

Screenshot 2021-02-01 at 07 47 29

```diff
--- a/liwords-ui/src/lobby/register.tsx
+++ b/liwords-ui/src/lobby/register.tsx
@@ -103,6 +103,10 @@ export const Register = () => {
                   required: true,
                   message: 'Please input your username',
                 },
+                {
+                  min: 3,
+                  message: 'Min username length is 3',
+                },
                 {
                   max: 20,
                   message: 'Max username length is 20',
```

jvc56: Good catch. Added.

domino14 (approved):

PR #417, opened 2021-02-09
Show editable avatar on Profile page

slipkin:
Video of system in action:
https://drive.google.com/file/d/10Fja6Zd6dSLnkZbe2K5_38TRwuKQYn_O/view

Add real, editable, DB-based avatar images.
Avatars show on Profile page (even for bots) and on game screens
Hide 'Edit' button on bios that aren't yours.
Client side no longer fakes bot avatars.

Integration notes:
* Current main.go disables editing. 
* To enable editing, update line 147 with your choice of uploader, e.g.:
  	profileService := pkguser.NewProfileService(stores.UserStore, pkguser.NewXTUploadService())
* The XTUploadService works only with SMALL images (~3K bytes), but it functions if you want to see the system in action.

Future:
antd-img-crop looks like it would be a nice additional, but I haven't been able to install it properly yet.


pkg/entity/user.go
```diff
@@ -130,3 +131,11 @@ func (u *User) RealName() string {
 	}
 	return ""
 }
+
+func (u *User) AvatarUrl() string {
+	if u.IsBot {
+		return "https://woogles-prod-assets.s3.amazonaws.com/macondog.png"
```
domino14, line 137: this should probably be a temporary thing. I can upload the macondog avatar by signing in as HastyBot as well .. and perhaps other bots can have other avatars

slipkin: I added one commit to use the macondog only if the bot has no avatar_url in the DB.

domino14:
This works; need a way to delete old avatars from S3 when uploading a new one.

domino14 (approved):

liwords-ui/src/shared/player_avatar.tsx
```diff
@@ -1,16 +1,114 @@
...
+  const [avatarUrl, setAvatarUrl] = useState<string | undefined>('');
+  const [avatarFile, setAvatarFile] = useState(new File([''], ''));
+
+  useEffect(() => {
+    setAvatarUrl(props.player?.avatar_url);
+  }, [props.player]);
+
+  useEffect(() => {
+    setAvatarErr('');
+    var fileInput = document.getElementById(
```
domino14, line 32: you should use `let` instead of `var`

liwords-ui/src/shared/player_avatar.tsx
```diff
@@ -1,16 +1,114 @@
...
+      visible={updateModalVisible}
+      okText="Upload"
+      okButtonProps={{ disabled: okButtonDisabled }}
+      onCancel={() => {
+        setUpdateModalVisible(false);
+      }}
+      onOk={() => {
+        console.log(avatarFile);
+        var reader = new FileReader();
+        reader.onload = function () {
```
domino14, line 70: using an arrow function in general is a bit better `reader.onload = () => { ...}`

PR #440, opened 2021-02-26
Scrub the Scrubs

jvc56:


pkg/profile/profile.go
```diff
@@ -92,12 +112,20 @@ func (ps *ProfileService) GetUsersGameInfo(ctx context.Context, r *pb.UsersGameI
 	var infos []*pb.UserGameInfo
 
 	for _, uuid := range r.Uuids {
-		user, err := ps.userStore.GetByUUID(ctx, uuid)
-		if err == nil {
+		user, userStoreErr := ps.userStore.GetByUUID(ctx, uuid)
+		avatarUrl := user.AvatarUrl()
```
andy-k, line 116: remember to check `userStoreErr`, otherwise `user` may be `nil` and this may panic

jvc56: Good catch, I believe I have fixed it.

pkg/tournament/service.go
```diff
@@ -451,41 +494,38 @@ func authenticateDirector(ctx context.Context, ts *TournamentService, id string,
...
+		}
 
-func (ts *TournamentService) GetRecentClubSessions(ctx context.Context, req *pb.RecentClubSessionsRequest) (*pb.ClubSessionsResponse, error) {
-	return ts.tournamentStore.GetRecentClubSessions(ctx, req.Id, int(req.Count), int(req.Offset))
+		_, known = knownUsers[playerTwo]
+		if !known {
+			knownUsers[playerTwo] = mod.IsCensorable(ctx, us, playerTwo)
+		}
+		if knownUsers[playerTwo] {
+			game.Players[0].Username = mod.CensoredUsername
```
domino14, line 525: should this index be a 1?

jvc56: Yes, that's my mistake.

PR #451, opened 2021-03-09
Misc

domino14:


README.md
```diff
@@ -29,6 +29,8 @@ This source code is AGPL-licensed. You can modify the source for this app, or fo
 127.0.0.1	liwords.localhost
 ```
 
+(If you are on Windows and you want to use Chrome, you cannot use `.localhost`. Use `liwords.local` in your `C:\Windows\System32\drivers\etc\hosts`.)
```
ddugovic, line 32: I much appreciate your help; the error I encountered was due to my strange configuration (liwords is running on a separate machine than Windows/Chrome trying to access it).  Regardless I like the liwords.local suggestion.

PR #452, opened 2021-03-11
Logistic Win Boost

jvc56:


domino14:
has this been data reviewed?

jvc56:
No, not yet, at least not that I've heard over. Pull this branch and run the ratings tests to see them.

domino14:
it seems reasonable to me. it does seem to compress ratings a bit more though.. (1964 to 612 with only 5 players over 1900)

jvc56:
I don't think we should use live data for these range measurements. None of those results have matched anything we've seen on Woogles. The settings are just too different, it should only be used as a sanity check. The "go for the win" tables are going to have more value here.

PR #463, opened 2021-03-17
translate registration error

andy-k:


domino14:
This should work, but I think it pushes the problem down the road of having a proper translation layer for these errors. Go errors should always look lowercase with no final punctuation.

Still, we're far away from having some frontend translation layer for this, so this should be fine to merge. Thanks for finding this!

PR #466, opened 2021-03-18
Add admin widgets for running tournaments

domino14:
prior to merging in entire Josh tournament branch (#406) we want to run HOPPY and likely Can-Am using tournament mode. We need to make admin widgets for some of the most tricky actions (re-pairing people, re-pairing rounds, assigning byes / forfeits for people who left early, etc)


domino14:
Some widgets complete, added Check In button. 

Add more widgets later.

PR #480, opened 2021-03-20
2021 02 22 settings screen

slipkin:
Hey folks!

Here's the Settings screen PR.

Here's what I added since I sent the video:
=======
* Close Account: renamed "Don't close" button
* Added "Forgot password" link on the Change Password pane
* Change-password error should go directly under left field, right where the error about a blank field goes.
* Avatar uploading has no more modal dialog; the user just clicks Change (or Add a Profile Photo) and chooses a file.
* Close Account: The confirmation button calls the new endpoint, shows an error on failure, and shows the Closed Account page on success. (Closed Account is a little different from the mocks because the back end currently just sends an email, rather than actually closing the account).
=======

There are a few integrations that are hard for me to complete, and I've listed them below. I'd love to hand this off as-is if possible, but mostly because I think you can land it faster than I can because of the listed integrations. It would be trivial for me to back off some of the settings (e.g. Language support) if that's helpful.

The loose ends are:

* Score and clock placement
  Is there any support for this? I don't know what underlying setting(s) to change, or if they exist. Does the site currently support this? I couldn't find it.

* Language support
  I've removed the "Enable All Lexicons" control (and all the other top-bar controls!) on the Profile page in favor of the finer-grained language (CSW/NWL/Polish/Norwegian) controls on the Settings page. However, those new controls aren't hooked up to anything.

* Mobile support
  I did not make the Settings page work with mobile. There is a "categories" div at the left (the list of panes available in Settings) and a "category" div at the right (the currently selected pane). I think(?) they just need @media settings for the narrower widths, but I'm not sure. These top-level items do not use Ant's Row/Col elements (although lower-level components do). I made an attempt to match the "columns" on the main screen (chat/games/announcements) but I don't know how well I did.

* The "Contribute" link takes the user to the About page; I didn't want to expose any still-private links there.

* I didn't add the "Secret" features from the mocks

* Blocked player: I didn't add the rollover effect on the "X"

* Profile page: It would be good to do something better than "show cloud error" when user doesn't exist. (This isn't new to this CL, but I noticed it during development. The Settings page will redirect to the home page in this situation).


pkg/profile/profile.go
```diff
@@ -77,6 +77,7 @@ func (ps *ProfileService) GetProfile(ctx context.Context, r *pb.ProfileRequest)
 	return &pb.ProfileResponse{
 		FirstName:       user.Profile.FirstName,
 		LastName:        user.Profile.LastName,
+		FullName:		 user.RealName(),
```
domino14, line 80: You should use `gofmt` on Go src files

domino14:
This is awesome. To answer a couple of the questions:

- The site does not currently support moving clock / scorecard locations. We can probably add this later.
- Is it possible to save people's lexicon settings (and perhaps even dark/light settings) inside their profile somewhere? Right now, the `enableAllLexicons` and `enableECWL` etc just sets `localStorage` settings. I think though that these should likely be persisted on the backend, especially as we add more settings.
- I believe the Patreon link is no longer private? I will check (do you know @atomiccupcake ?)

domino14:
Please run prettier on the tsx files as well (I believe you can make your code editor re-run it automatically on save)

slipkin:
Hmm, I do have prettier set up. Can you give an example of a bad .tsx file?
Thanks!

domino14:
> Hmm, I do have prettier set up. Can you give an example of a bad .tsx file? Thanks!

For example, `liwords-ui/src/settings/blocked_players.tsx` uses tabs instead of spaces.

slipkin:
Gotcha! Try now, please.

In Sublime Text, a bunch of files were marked as using TabSize = 2 (or 4)
rather than Spaces=2, and apparently Prettier respects that setting.

Let me know if I missed any, thanks!

Seth

atomiccupcake (changes requested):
A few things.

twirp/user_service.ProfileService/GetProfile doesn't seem to exist, so I get an error when I go to the settings page.

All buttons should read like "No, keep my account" but CSS is overriding to "No, Keep My Account". This is a change that Conrad made, but the CSS wasn't updated yet because we've yet to have any multi word buttons. Remove line 638 in base.scss to correct and future proof?
As it is you have both styles next to each other on the page.

In dark mode, I can't read the confirmation text on the close account checkbox.
image

Toggles overlap on smaller screen sizes
image
You can fix this by 
`.toggle-section > div {
    padding-right: 60px;
}` That way those title and description bits will always leave room on the side for the toggle.

Until we can set the languages, the only think will work for "enable all lexicons" is allowing them to show up for new match requests, as hooked up before to local storage. The text Conrad wrote doesn't work (it doesn't effect the seeks you see) and those languages don't exist. So pull that out for now and just put the toggles we had there for now (or comment out the language section completely for now until we can build that)

The only other thing that I would like, is on the profile page (your own only), where all of this crap used to live, can we put a "Edit settings" link that goes here? (Just because existing users and people following Doug's how to and Will's videos will look there first, and a link will bring them to the right place.)

jvc56:
Do you mean '/twirp/user_service.ProfileService/GetPersonalInfo' doesn't exist? I think there is a 'GetProfile'.

atomiccupcake:
> Do you mean '/twirp/user_service.ProfileService/GetPersonalInfo' doesn't exist? I think there is a 'GetProfile'.

Yes 
image
When I hit the settings page

andy-k:
you'll need to `docker-compose restart app`, because [the GetPersonalInfo endpoint is new in this same PR](https://github.com/domino14/liwords/pull/480/files#diff-8fba28f2b191eaf308890dbc3ffad8e55b5448849e842be536db0cdddaa12c6eR180)

atomiccupcake:
Looks great! One last thing. Show tiles remaining doesn't seem to be hooked up to anything and should be hidden until it is.

(If you want to hook it up,  I'd do it in a different branch, you'd want the pool card when that's OFF to only show the count total (17 tiles remaining) and nothing else, and we'd want the default setting to be on to SHOW tiles remaining. So your flag would actually be hideRemainingTiles and the toggle would default to on. But again, separate project for people that want to track manually or whatever.)

atomiccupcake (approved):

slipkin:
🎉 thanks!

PR #485, opened 2021-03-28
Send email on mod action

jvc56:


domino14:
waiting on tos?

jvc56:
An important note about this feature. If the case insensitive string "cheat" appears in the note of the action, it will be treated as a cheating violation and a different email will be sent.

jvc56:
Ready for merge unless we want to email the user the moderator's note.

PR #487, opened 2021-03-29
adjust time low cutoff

andy-k:
to 1/5 of initial time, or 30sec, whichever is higher.

so instead of at 3min, the clock turns yellow at
- for 2min games or shorter, at 30sec
- for 3min games, at 36sec
- for 4min games, at 48sec
- for 15min games, at 3min
- for 25min games, at 5min

a 1/4+5 game will start in the yellow region, but can reach green region by accumulating more than 30sec.

### DIFF
--- a/liwords-ui/src/gameroom/player_cards.tsx
+++ b/liwords-ui/src/gameroom/player_cards.tsx
@@ -49,7 +49,7 @@ const PlayerCard = React.memo((props: CardProps) => {
   const timeStr =
     isExamining || props.playing ? millisToTimeStr(props.time) : '--:--';
   // TODO: what we consider low time likely be set somewhere and not a magic number
-  const timeLowCutoff = props.initialTimeSeconds / 5;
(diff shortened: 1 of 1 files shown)


liwords-ui/src/gameroom/player_cards.tsx
```diff
@@ -48,7 +49,8 @@ const PlayerCard = React.memo((props: CardProps) => {
   const timeStr =
     isExamining || props.playing ? millisToTimeStr(props.time) : '--:--';
   // TODO: what we consider low time likely be set somewhere and not a magic number
-  const timeLow = props.time <= 180000 && props.time > 0;
+  const timeLowCutoff = Math.max(props.initialTimeSeconds / 5, 30000);
```
domino14, line 52: How were these arrived at? It seems that having a 3 minute timer would result in for example an arbitrary-looking cutoff of 36 seconds, 4 minutes would be 48 seconds, etc. It's a bit strange. I think if there was some rounding to the nearest 15 or 30 seconds it might be better.

PR #495, opened 2021-04-01
Blindfold Mode

jvc56:


liwords-ui/src/utils/cwgame/blindfold.ts
```diff
@@ -0,0 +1,466 @@
...
+
+export const parseBlindfoldCoordinates = (
+  coordinates: string) : BlindfoldCoordinates | undefined => {
+  if (!validBlindfoldCoordinates.has(coordinates)) {
+    return undefined
+  }
+  return validBlindfoldCoordinates.get(coordinates)
+}
+
+const validBlindfoldCoordinates = new Map([['1A', {row: 0, col: 0, horizontal: true} ],
```
jvc56, line 17: I will do this the right way using regexes instead of this big map.

PR #502, opened 2021-04-08
pb experiment

domino14:
a lot of our XHRs are JSON and we define redundant JSON types to handle these. We should keep these as protobuf as to not maintain two separate representations of the same data. Downstream from the data transfer code we can have JSON objects with what we need, but the XHRs shouldn't be handling JSON and the sockets handling PB with same but different types


domino14:
This sucks. For this message:

```proto
message TournamentMetadataResponse {
  string name = 1;
  string description = 2;
  repeated string directors = 3;
  string slug = 4;
  string id = 5;
  TType type = 6;
  repeated string divisions = 7;
}
```

`protoc-gen-ts` generates an AsObject that looks like this:

```typescript
export namespace TournamentMetadataResponse {
  export type AsObject = {
    name: string,
    description: string,
    directorsList: Array<string>,
    slug: string,
    id: string,
    type: TTypeMap[keyof TTypeMap],
    divisionsList: Array<string>,
  }
}

```

Notice that it adds a `List` suffix to lists. This means that I can't just use `TournamentMetadataResponse.AsObject` as a type  (for example here: `axios.post<TournamentMetadataResponse.AsObject>(`) -- the backend will return the correct `directors` and `divisions` and not `directorsList` and `divisionsList`. I don't know how to fix this; I don't think there's a good alternative to `protoc-gen-ts`.

One thought is to use OpenAPI and the proto files and move away from Twirp (which is not good for what we're doing anyway, we should be trying to be RESTful with our API, since we have a lot of GET requests and can probably do things like ETags and caching etc, but Twirp makes everything a POST). Perhaps OpenAPI will generate the right typescript files with the right types?

domino14:
this is broken, closing for now

PR #504, opened 2021-04-08
Automod

jvc56:


pkg/stores/user/db.go
```diff
@@ -317,6 +322,7 @@ func (s *DBStore) toDBObj(u *entity.User) (*User, error) {
 		IsAdmin:     u.IsAdmin,
 		IsDirector:  u.IsDirector,
 		IsMod:       u.IsMod,
+		Notoriety:   u.Notoriety,
```
andy-k, line 325: with this, it may be impossible to .Set a user's notoriety back to zero.

jvc56: Why might it be impossible?

andy-k: zero values trap.

jvc56: oh, that dumb thing again, wouldn't this also apply to unmodding and unadminning people?

andy-k: [yes. `.Set` cannot be used for that purpose.](https://github.com/domino14/liwords/pull/535) 😅

jvc56: Added SetNotoriety to avoid this trap.

pkg/stores/user/cache.go
```diff
@@ -276,3 +276,9 @@ func (c *Cache) Set(ctx context.Context, u *entity.User) error {
 	c.cache.Add(u.UUID, u)
 	return nil
 }
+
+// This was written to avoid the zero value trap
+func (c *Cache) SetNotoriety(ctx context.Context, u *entity.User, notoriety int) error {
+	u.Notoriety = notoriety
+	return c.backing.SetNotoriety(ctx, u, notoriety)
```
andy-k, line 283: i know the other methods[1] have the same problem too, but (aside from not being able to transact[2]) updating the cached value before the db returned success seems incorrect. the slightly more correct version would be something like
```go
err := c.backing.SetNotoriety(ctx, u, notoriety)
if err != nil {
  return err
}
u.Notoriety = notoriety
return nil
```
[1] in particular `.Set` itself, by design.
[2] with transaction, this becomes even less clear, because the cache should be updated on successful commit, not on a successful backing.SetNotoriety.

pkg/mod/automod.go
```diff
@@ -0,0 +1,189 @@
...
+		}
+	} else if user.Notoriety > 0 {
+		user.Notoriety -= NotorietyDecrement
+		if user.Notoriety < 0 {
+			user.Notoriety = 0
+		}
+	}
+
+	if previousNotorietyScore != user.Notoriety {
+		return us.Set(ctx, user)
```
andy-k, line 182: this should probably use `.SetNotoriety`

domino14:
- Add a punishment for confirming a nudge and/or denying a cancel and still not making a play

pkg/stores/mod/db.go
```diff
@@ -0,0 +1,76 @@
+package mod
+
+import (
+	ms "github.com/domino14/liwords/rpc/api/proto/mod_service"
+
+	"github.com/jinzhu/gorm"
```
domino14, line 6: i know we have this all over the code but `"gorm.io/gorm"` is the new one. I don't think it's worth changing this at the moment since we have this everywhere.

PR #512, opened 2021-04-13
Automatically Delete Account

jvc56:


pkg/stores/user/db.go
```diff
@@ -849,6 +849,42 @@ func (s *DBStore) ResetStatsAndRatings(ctx context.Context, uid string) error {
...
+	if result := s.db.Model(p).Update("about", ""); result.Error != nil {
+		return result.Error
+	}
+	if result := s.db.Model(p).Update("title", ""); result.Error != nil {
+		return result.Error
+	}
+	if result := s.db.Model(p).Update("avatar_url", ""); result.Error != nil {
+		return result.Error
+	}
+	return s.db.Model(p).Update("country_code", "").Error
```
domino14, line 877: can we make this all one single Update query?

jvc56: perhaps

jvc56: yes

domino14 (approved):

Issue #513, opened 2021-04-16
Need endpoint that returns moderators and admins


atomiccupcake:
So that we can tag people in the UX. Similar in format and functionality to the "GetFullBlocks" endpoint.

Response format:

{
  admin_user_ids: [],
  mod_user_ids: [],
}

magratheazaphod:
some additional factors on the engineering implementation:
* These look like SELECT * queries from our user database. However, if we do this the naive way, we'll get 1000s of queries an hour, which is bad.
* Therefore, we want some kind of caching layer that updates who admins/mods are periodically.

domino14:
We ended up using queries with indices, and caching on the frontend. I think this is fast enough.

@atomiccupcake  you can test with 

```
curl -v 'http://liwords.localhost/twirp/user_service.SocializeService/GetModList' \
  -H 'Content-type: application/json' -d '{}'
```

PR #519, opened 2021-04-16
add moderator list endpoint

andy-k:
#513 

```
curl -v 'http://liwords.localhost/twirp/user_service.SocializeService/GetModList' \
  -H 'Content-type: application/json' -d '{}'
```


pkg/stores/user/db.go
```diff
@@ -886,3 +886,30 @@ func (s *DBStore) Count(ctx context.Context) (int64, error) {
 func (s *DBStore) CachedCount(ctx context.Context) int {
 	return 0
 }
+
+func (s *DBStore) GetModList(ctx context.Context) (*pb.GetModListResponse, error) {
+	var users []User
+	if result := s.db.
+		Where("is_admin = ?", true).Or("is_mod = ?", true).
```
domino14, line 893: I think we should add an index on these fields. I can do it manually on prod but we can change these lines up top to chain in some admin/mod indices? 

```go
	db.Model(&User{}).
		AddUniqueIndex("username_idx", "lower(username)").
		AddUniqueIndex("email_idx", "lower(email)").
		AddIndex("api_key_idx", "api_key")
```

domino14 (changes requested):
just a comment on indices; the rest looks great

pkg/stores/user/db.go
andy-k: added some indexes, but they don't seem to be used. feel free to revert that commit.

[partial indexes](https://www.postgresql.org/docs/current/indexes-partial.html) may be more appropriate. but, not sure how to express that in gorm.

Issue #515, opened 2021-04-16
Need more information in chat message


atomiccupcake:
To implement the UX, we need more information in the messages:

Current response:
{
  channel: "chat.tournament.dallasbedford"
  id: "1618534097639-0"
  message: "Okay Caesar, you initiate"
  timestamp: 1618534097639
  userId: "Eb67dVRHCQVmiKsYxbXZbi"
  username: "afeicher"
}

Proposed response:
{
  country_code: "us"
  avatar_url: "https://...."
  channel: "chat.tournament.dallasbedford"
  id: "1618534097639-0"
  message: "Okay Caesar, you initiate"
  timestamp: 1618534097639
  userId: "Eb67dVRHCQVmiKsYxbXZbi"
  username: "afeicher"
}

atomiccupcake:
Current handler for MessageType.CHAT_MESSAGE starts on like 382 of socket_handler.ts

domino14:
@atomiccupcake Every chat message now has an avatar_url and a country_code, whether it comes from the socket or through XHR. Thank you @andy-k!

Issue #518, opened 2021-04-16
Add avatar and flag to seek message


atomiccupcake:
Please add the country code and avatar of the  user seeking a game (or making a match request)

Issue #517, opened 2021-04-16
Add avatar and flag to active game message


atomiccupcake:
Please include the country code and avatar of each user in the players for the active games, so we can display them in the watch list

PR #523, opened 2021-04-16
add some profile info in chat

andy-k:
#515 (inefficient implementation, need help)


andy-k:
would it work better to have a new "bulk get profiles by user ids" endpoint instead? so #515 #517 #518 API unchanged, and frontend consolidates requests for user profiles across chats/games/seeks (and cache them in global store) and render the flag and avatar when that endpoint returns. (so chat from a new user will first come with no user profile, and then user profile comes and is used for all chats by that user.)

andy-k:
(marking this as draft. this works, but can be closed if #524 is merged instead.)

andy-k:
this is now using the efficient endpoint in #524. but using this will complicate the frontend code, and will only be a win iff chatters only chat and don't play games. after merge, we can revert it if it turns out to be unused.

domino14 (approved):

Issue #516, opened 2021-04-16
Need database record and endpoints for friends


atomiccupcake:
This should work almost exactly like blocks.

Please make a table for keeping track of friends.
Make an endpoint that returns the friends you have selected.

Response body:
{
  friendsList:
    [
      {
         user_id: "dztWVeEgY8f9k4DP6oy6MF", 
        username: "Lola",
        avatar_url: "https://...",
        country_code: "us",
     } ,
  ],
}

Make an endpoint for adding a friend. Request body:
   {user_id: "dztWVeEgY8f9k4DP6oy6MF"} 
Make an endpoint for removing a friend. request body:
  {user_id: "dztWVeEgY8f9k4DP6oy6MF"}

domino14:
@atomiccupcake - this was implemented by Andy. See description of this PR:

https://github.com/domino14/liwords/pull/525

The API is very similar to that of blocks. We don't have avatar/country code in the response yet. This will likely come in another PR.

PR #524, opened 2021-04-16
get brief profiles

### LINKED ISSUE #515: Need more information in chat message

andy-k:
this implements my suggestion in https://github.com/domino14/liwords/pull/523#issuecomment-820950865
- this adds a new hook (`useBriefProfile`), which takes a userid and returns the country code and avatar url
- it is ok to use this hook in many components, it will just issue one query for all
- it is ok to use the same user's brief profile multiple times (for example a ShowFlag and a ShowAvatar component, or using UsernameWithContext for every chat entry), it will deduplicate
- the hook will rerender the whole screen when it loads something
- this would resolve issues like #515, #517, #518, and make #525 work for #516
- #523 is no longer needed

usage: refer to `UsernameWithContext` for example. (the code is false'd out so there's no ui change)
- create a component to render the user
- `const briefProfile = useBriefProfile(props.userID);`
- `briefProfile` can be undefined (loading), null (not found), or an object with getCountryCode() and getAvatarUrl()
- the component will first render with undefined. don't crash if `briefProfile` isn't ready yet, it is ok to return null

current implementation notes:
- frontend never deallocates loaded objects, but the cache gets reset along with resetStore
- frontend remembers the last time the object was loaded, and will reload if the current version is older than 5 minutes ago
- backend does not cache


domino14 (changes requested):
I like what you have done with front-end caching. However, I think we may need a more hybrid solution.

For example, if people are in a HOPPY type tournament and someone new sends a message, then that's 100+ simultaneous XHRs and db queries. I wonder if there's an in-between way to handle this. Adding the avatars and flags directly to the chat message like you did in your other PR also makes sense and results in fewer per-user queries but also seems inefficient. Can we think of other ideas?

andy-k:
added some backend cache.

atomiccupcake (approved):

domino14 (approved):

Issue #528, opened 2021-04-17
Opt out of the evil empire


ddugovic:
https://twitter.com/ornicar/status/1382960243861291010
https://www.eff.org/deeplinks/2021/03/googles-floc-terrible-idea

ddugovic:
Maybe also don't let Google track users who read documents such as the Quick Start Guide (Google Drive document).

PR #529, opened 2021-04-17
deny floc locally

andy-k:
#528 for local.

production has to be patched separately


domino14:
how do i patch this on prod? i think via cloudfront?

PR #530, opened 2021-04-18
Tag messages in chat from mods and admins

atomiccupcake:
image

### DIFF
--- a/liwords-ui/src/App.tsx
+++ b/liwords-ui/src/App.tsx
@@ -11,6 +11,7 @@ import { Lobby } from './lobby/lobby';
 import {
   useExcludedPlayersStoreContext,
   useResetStoreContext,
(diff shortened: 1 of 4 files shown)


liwords-ui/src/App.tsx
```diff
@@ -96,6 +109,29 @@ const App = React.memo(() => {
     }
   }, [getFullBlocks, pendingBlockRefresh]);
 
+  const getMods = useCallback(() => {
+    axios
+      .post<ModsResponse>(
+        toAPIUrl('user_service.SocializeService', 'GetModList'),
+        {},
+        { withCredentials: true }
```
domino14, line 117: this particular call does not require credentials (so that anon users can also see who's a moderator/admin).

PR #532, opened 2021-04-18
Notepad clear

atomiccupcake:
Allows notepad to be quickly reset, compatible with catcam easter egg
image
image
image

### DIFF
--- a/liwords-ui/src/gameroom/notepad.tsx
+++ b/liwords-ui/src/gameroom/notepad.tsx
@@ -6,7 +6,7 @@ import React, {
   useMemo,
 } from 'react';
 import { Button, Card } from 'antd';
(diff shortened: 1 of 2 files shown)


liwords-ui/src/gameroom/notepad.tsx
```diff
@@ -135,6 +135,30 @@ export const Notepad = React.memo((props: NotepadProps) => {
     }
     numWolgesWas.current = numWolges;
   }, [numWolges]);
+  const controls = useCallback(
```
andy-k, line 138: - how about `useMemo` (and use `controls` instead of `controls()` at the two places below)?
- would it be desirable to also set focus on the emptied notepad (`notepadEl.current?.focus();`)?

Issue #514, opened 2021-04-16
Presences need to be global


atomiccupcake:
Presences need to return all users on the site, regardless of location. Should also return active games. (Could temporarily be a boolean that just says: "inGame: true",

Current response:
{username: "Andrea518", userId: "FSqg3yi3DFSCJpyeqR5fK4", channel: "chat.lobby", isAnonymous: false, deleting: true}

Proposed response:
{
  activeGames: {
    omgWords: ["G5aWaZ7K"],
    meowWords: ["Y2zZ43a", "2b43FFa"], 
   }.
  inGame: true,  //implement this only if activeGames is too hard
  username: "Andrea518",
  userId: "FSqg3yi3DFSCJpyeqR5fK4",
  channels: ["chat.lobby"]
  isAnonymous: false,
  deleting: true
 }

atomiccupcake:
Current presence handlers are on lines 412 through 449 of socket_handlers.ts Message.USER_PRESENCE and Message.USER_PRESENCES

PR #541, opened 2021-04-21
add presence endpoint

andy-k:
#514 

this does not change Message.USER_PRESENCE and Message.USER_PRESENCES, but enhances #525 (which implements #516) as follows*

`user_service.SocializeService/GetFollows` `{}` now returns `{"users":[{"uuid":"...","username":"...","channels":["...","..."]}]}`

so now you can really ~stalk~ follow those who you are following. to follow others, you need to follow them first

\* pun intended


andy-k:
#552 should make this unnecessary.

PR #549, opened 2021-04-24
Chat Rate Limiter

jvc56:
Comrade RightBehindYou requests that you give him the power to silence the will of the people on wolges.

### DIFF
--- a/pkg/bus/bus.go
+++ b/pkg/bus/bus.go
@@ -29,7 +29,8 @@ import (
 )
 
 const (
-	MaxMessageLength = 500
+	MaxMessageLength    = 500
+	MessageCooldownTime = 5
 
 	AdjudicateInterval   = 10 * time.Second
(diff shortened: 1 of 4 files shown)


jvc56:
Architecture needs to be rethought.

domino14:
I would use a Redis Hash key: https://redis.io/topics/data-types-intro

It can expire after some time (a few hours?). You can do this with an `EXPIRE keyname` command after you use it.

The HASH would have the last chat string and the last timestamp as two separate fields. 

I might want to write a Lua script. It could be called something like `AddChat` and it would return an optional error. The error can be propagated to the frontend if it's the right error (`rate-limit`, etc). Inside the script you can atomically check to see if the chat matches the last chat, if the timestamp is too frequent, etc.

domino14:
reimplemented in #578 -- thank you!

PR #554, opened 2021-04-26
Allow users to crop their avatars

atomiccupcake:
image

### DIFF
--- a/liwords-ui/src/settings/avatar_cropper.tsx
+++ b/liwords-ui/src/settings/avatar_cropper.tsx
@@ -39,7 +39,7 @@ export const AvatarCropper = React.memo((props: Props) => {
         canvas.width,
         canvas.height
       );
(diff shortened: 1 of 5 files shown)


liwords-ui/src/settings/avatar_cropper.tsx
```diff
@@ -0,0 +1,131 @@
...
+        croppedArea.x,
+        croppedArea.y,
+        croppedArea.width,
+        croppedArea.height,
+        0,
+        0,
+        canvas.width,
+        canvas.height
+      );
+      const newDataUrl = canvas.toDataURL('/image.jpeg', 1);
```
andy-k, line 42: `'/image.jpeg'` in two places (previous code used `'image/jpeg'`)

atomiccupcake: Fixed! Thanks

PR #560, opened 2021-04-26
call api only after logging in

andy-k:
doesn't make sense to call GetFullBlocks before logging in only to have it return 401


atomiccupcake:
This pr prevents logged out users from seeing chat. Would be nice if there were a way to setExcludedPlayersFetched(true) for them, which the 401 accomplished, though clumsily.

PR #430, opened 2021-02-16
update ECWL gaddag

andy-k:

PR #561, opened 2021-04-26
update ecwl

andy-k:
https://github.com/Fj00/CWL/tree/0e6290ac38b667e40c78fb3fee6ee3243fef9b7c

supersedes #430


domino14:
Could you remove the `enableECWL` localStorage setting? (we still want to hide ECWL by default, but only with the `showAllLexicons` toggle)

PR #562, opened 2021-04-27
Lots of tile colors, previewer on preference page

atomiccupcake:
11 choices for tiles (in addition to the default). The Charcoal one is the one that will be hardcoded for NASSC.

image
image
image

There will be something similar for board, with NASSC's hardcoded design probably not available for everyone to select.

### DIFF
--- a/liwords-ui/src/gameroom/scss/gameroom.scss
+++ b/liwords-ui/src/gameroom/scss/gameroom.scss
@@ -714,7 +714,7 @@ p.streak-loss {
       }
       @include userTile() {
         color: ut($tile-blank-text); // Fallback if border is a gradient
-        color: ut($tile-last-background); // Fallback if border is a gradient
+        color: ut($tile-last-background);
         background: ut($tile-last-blank);
       }
     }
@@ -750,11 +750,11 @@ p.streak-loss {
   }
   &.blank p.rune {
     @include colorModed() {
-      background-color: m($tile-background-quartenary);
+      background-color: m($tile-background-quaternary);
       color: m($tile-background);
     }
     @include userTile() {
-      background-color: ut($tile-background-quartenary);
+      background-color: ut($tile-background-quaternary);
       color: ut($tile-blank-text);
     }
   }
--- a/liwords-ui/src/gameroom/scss/gameroom.scss
+++ b/liwords-ui/src/gameroom/scss/gameroom.scss
@@ -1,5 +1,6 @@
 @import '../../base';
 @import '../../shared/layout';
+@import '../../tile_modes';
 
 @include button;
 @include notification;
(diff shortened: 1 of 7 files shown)


domino14:
wow!

liwords-ui/src/settings/preferences.tsx
```diff
@@ -28,12 +29,66 @@ const KNOWN_TILE_ORDERS = [
...
+    name: 'Teal',
+    value: 'tealish',
+  },
+  {
+    name: 'Pastel',
+    value: 'pastel',
+  },
+  {
+    name: 'Fuschia',
+    value: 'fuschia',
```
andy-k, line 71: i believe it's spelt fuCHSia

liwords-ui/src/tile_modes.scss
```diff
@@ -0,0 +1,157 @@
...
+            color-tile-text: #5c455f,
+            color-tile-last-background: #674f69,
+            color-tile-last-text: #fad4c9,
+            color-tile-last-blank: #fad4c9,
+        ),
+        fuschia: (
+          color-tile-background: #c82559,
+          color-tile-background-secondary: #ef5284, //Circle in blank
+          color-tile-background-tertiary: #940e39, //Tentative background
+          color-tile-background-quartenary: #c82559,  //Tentative blank circle
```
andy-k, line 121: (also good time to fix quaternary spelling?)

liwords-ui/src/gameroom/scss/gameroom.scss
```diff
@@ -686,43 +701,62 @@ p.streak-loss {
...
   &.blank {
     p.rune {
       @include colorModed() {
         background-color: m($tile-last-blank);
         color: m($tile-last-background);
-
+      }
+      @include userTile() {
+        color: ut($tile-blank-text); // Fallback if border is a gradient
+        color: ut($tile-last-background); // Fallback if border is a gradient
```
domino14, line 717: does this concatenate both `color` settings? Not sure how this works in SCSS.

atomiccupcake: In CSS only the last  valid style counts. In this case, if the tile-last-background is a gradient or picture or something else that's not valid for text (or border), it will use the one before. Metallic is special, but when we do Kickstarter colorways we may even have backgrounds that are images. In which case fallback. I'll remove that second comment. The first one is the fallback.

liwords-ui/src/settings/preferences.tsx
```diff
@@ -28,12 +29,66 @@ const KNOWN_TILE_ORDERS = [
...
+  },
+];
+
 export const Preferences = React.memo((props: Props) => {
   const { useState } = useMountedState();
 
   const [darkMode, setDarkMode] = useState(
     localStorage?.getItem('darkMode') === 'true'
   );
+  const initalTileStyle = localStorage?.getItem('userTile') || 'default';
```
domino14, line 89: this should also be `initialTileStyle`

atomiccupcake: Sigh

PR #564, opened 2021-04-27
Revert some blindfold changes and adding passing and challenging as keyboard options

jvc56:


liwords-ui/src/gameroom/board_panel.tsx
```diff
@@ -834,6 +834,32 @@ export const BoardPanel = React.memo((props: Props) => {
...
+          } else if (
+            blindfoldCommand.toUpperCase() === 'CHAL' &&
+            !props.gameDone
+          ) {
+            setCurrentMode('NORMAL');
+            evt.preventDefault();
+            if (handleChallengeShortcut.current)
+              handleChallengeShortcut.current();
+            say('Press Enter to confirm challenge.', '');
+            return;
```
jvc56, line 862: I cannot figure out how to execute challenge or passes here. Preferably would not need confirmation since the user already typed out four characters.

jvc56: I figured it out.

PR #566, opened 2021-04-28
add flag outline

andy-k:
currently 🇯🇵 looks like doing a live "recording..." on light background.

Screenshot 2021-04-27 at 08 10 08


domino14:
#574

PR #578, opened 2021-05-02
limit chat rate

andy-k:
reimplement #549

### DIFF
--- a/pkg/bus/chat.go
+++ b/pkg/bus/chat.go
@@ -38,13 +38,10 @@ func (b *Bus) chat(ctx context.Context, userID string, evt *pb.ChatMessage) erro
 
 	// Regulate chat only if the user is not privileged and the
 	// chat is not a game chat
-	regulateChat := "regulated"
-	if sendingUser.IsAdmin ||
+	regulateChat := !(sendingUser.IsAdmin ||
 		sendingUser.IsMod ||
 		sendingUser.IsDirector ||
-		strings.HasPrefix(evt.Channel, "chat.game.") {
-		regulateChat = "unregulated"
-	}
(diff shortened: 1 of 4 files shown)


pkg/bus/chat.go
```diff
@@ -36,6 +36,16 @@ func (b *Bus) chat(ctx context.Context, userID string, evt *pb.ChatMessage) erro
 		return err
 	}
 
+	// Regulate chat only if the user is not privileged and the
+	// chat is not a game chat
+	regulateChat := "regulated"
```
jvc56, line 41: Can this (and the corresponding variable in the lua script) be a boolean?

andy-k: i first tried boolean, but i ran into problems i didn't understand

jvc56: hmm, maybe @domino14 can have a look, but not a huge deal either way.

pkg/bus/chat.go
```diff
@@ -36,6 +36,16 @@ func (b *Bus) chat(ctx context.Context, userID string, evt *pb.ChatMessage) erro
...
+	if sendingUser.IsAdmin ||
+		sendingUser.IsMod ||
+		sendingUser.IsDirector ||
```
domino14, line 44: this may be counterintuitive but tournament/club directors are not actually directors in the database (rather, they don't have to be). I can fix this though by eventually assigning them director privileges.

pkg/bus/chat.go
domino14: these could still be boolean, but the receiving function `AddChat` would take care of the actual implementation details (turn it into "regulated")

domino14 (approved):

PR #584, opened 2021-05-03
Friends please

atomiccupcake:
Adds the ability to add and remove friends to the user context menu.
Relabels "message" to "chat" in the user context menu, for clarity.
Makes players tab default for lobby.
Shows friends and their online state to the players tab.
Shows online players in your room (old presences) to the players tab.
Lets you search for players in the players tab.
Adds your friends list to the match screen options in the lobby if you haven't typed anything in the search field.
Adds present players to the match screen options in club/tournaments if you haven't typed anything in the search field.

image
image
image
image
image


liwords-ui/src/shared/follower.tsx
```diff
@@ -0,0 +1,74 @@
...
+        setPendingFriendsRefresh(true);
+      });
+  };
+
+  const DynamicTagName = (props.tagName ||
+    'span') as keyof JSX.IntrinsicElements;
+  return (
+    <DynamicTagName onClick={friendAction} className={props.className || ''}>
+      {friendText}
+    </DynamicTagName>
```
domino14, line 72: wow I had no idea you could do this!

liwords-ui/src/chat/players.tsx
```diff
@@ -0,0 +1,294 @@
...
+            return 1;
+          }
+          return 0;
+        })
+        .map(
+          (p): Partial<FriendUser> => {
+            return {
+              username: p.username,
+              uuid: p.uuid,
+              channel: new Array<string>().concat(p.channel),
```
andy-k, line 205: this would be `channel: [p.channel],` (turning a string to a string array, without first allocating an empty array; `concat` returns a new array).

but are the items in `PresenceEntity` all in the same channel? or are there potentially going to be duplicates if they are in multiple channels?

liwords-ui/src/chat/players.tsx
```diff
@@ -0,0 +1,294 @@
...
+        }
+      >
+        <section className="friends">
+          {loggedIn && <div className="breadcrumb">FRIENDS</div>}
+          {loggedIn &&
+            renderPlayerList(
+              filterPlayerListBySearch(searchText, Object.values(friends)),
+              'friends'
+            )}
+          {loggedIn && Object.values(friends).length === 0 && (
```
andy-k, line 268: (performance) each render calls `Object.values(friends)` several times and each call allocates a new copy. maybe `useMemo` it if we really need the values in friends values?
```js
const friendsValues = useMemo(() => Object.values(friends), [friends]);
```

liwords-ui/src/chat/players.tsx
```diff
@@ -0,0 +1,294 @@
...
+            </p>
+          )}
+        </section>
+        <section className="present">
+          {transformAndFilterPresences(presences, searchText).length > 0 && (
+            <div className="breadcrumb">
+              {getPresenceLabel(props.defaultChannelType || '')}
+            </div>
+          )}
+          {renderPlayerList(transformAndFilterPresences(presences, searchText))}
```
andy-k, line 281: (performance) `transformAndFilterPresences` is called twice, the first one is only to check length but it still does the work. this expression would benefit from `useMemo`.
```js
const transformedAndFilteredPresences = useMemo(
  () => transformAndFilterPresences(presences, searchText),
  [transformAndFilterPresences, presences, searchText]);
```

liwords-ui/src/lobby/seek_form.tsx
```diff
@@ -224,21 +233,29 @@ export const SeekForm = (props: Props) => {
...
+          }
+        )
+        .then((resp) => {
+          console.log('resp', resp.data);
+          const defaultOptions = tournamentID
+            ? presences.map((p) => p.username).filter((u) => u !== username)
+            : Object.values(friends).map((f) => f.username);
+          setUsernameOptions(
+            !searchText
+              ? defaultOptions || []
```
andy-k, line 252: (performance) `defaultOptions` is reevaluated on every `onUsernameSearch` (every keystroke), even when it isn't used. it should just be evaluated if `!searchText`.

liwords-ui/src/shared/follower.tsx
```diff
@@ -0,0 +1,74 @@
...
+  } else {
+    apiFunc = 'Add';
+    friendText = 'Add friend';
+    // Add some confirmation.
+  }
+
+  const friendAction = () => {
+    axios
+      .post(
+        toAPIUrl('user_service.SocializeService', `${apiFunc}Follow`),
```
andy-k, line 40: apiFunc should be the full text (AddFollow/RemoveFollow) so it's easier to grep for AddFollow next time

liwords-ui/src/shared/usernameWithContext.tsx
```diff
@@ -32,6 +35,22 @@ export const UsernameWithContext = (props: UsernameWithContextProps) => {
...
     <ul>
+      {loggedIn &&
+      !props.omitSendMessage &&
+      props.userID &&
+      props.userID !== userID &&
+      props.sendMessage ? (
+        <li
+          className="link plain"
+          onClick={() => {
+            if (props.sendMessage) {
```
andy-k, line 46: always true because this is checked outside

liwords-ui/src/store/socket_handlers.ts
```diff
@@ -458,7 +460,14 @@ export const useOnSocketMsg = () => {
 
             console.log('got presence entry', pe.toObject());
 
-            // TODO.
+            setFriends({
+              ...friends,
+              [pe.toObject().userId]: {
```
andy-k, line 465: (performance) this can be `[pe.getUserId()]:`. `pe.toObject()` would first allocate a new instance with all key-value pairs, and then we only need one field.

liwords-ui/src/App.tsx
```diff
@@ -159,6 +171,37 @@ const App = React.memo(() => {
...
+      .then((resp) => {
+        console.log('Fetched friends:', resp);
+        const friends: { [uuid: string]: FriendUser } = {};
+        Object.values(resp.data.users).forEach((f: FriendUser) => {
+          friends[f.uuid] = f;
+        });
+        setFriends(friends);
+      })
+      .catch((e) => {
+        console.log(e);
```
andy-k, line 190: it's calling this api when not logged in, and getting 401. (#565 patched the same issue for the blocked list)

liwords-ui/src/chat/chat.tsx
```diff
@@ -47,7 +48,7 @@ type JSONActiveChatChannels = {
   channels: Array<JSONChatChannel>;
 };
 
-export const Chat = React.memo((props: Props) => {
+export const Chat = (props: Props) => {
```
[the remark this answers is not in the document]
atomiccupcake: Putting it back. I was having trouble with it not rendering on some changes, but I straightened that out.

liwords-ui/src/chat/players.tsx
```diff
@@ -0,0 +1,294 @@
...
+  const [maxHeight, setMaxHeight] = useState<number | undefined>(0);
+  const [searchResults, setSearchResults] = useState<
+    Array<Partial<FriendUser>>
+  >([]);
+  const [searchText, setSearchText] = useState('');
+  const { presences } = usePresenceStoreContext();
+
+  const setHeight = useCallback(() => {
+    const tabPaneHeight = document.getElementById('chat')?.clientHeight;
+    setMaxHeight(tabPaneHeight ? tabPaneHeight - 117 : undefined);
```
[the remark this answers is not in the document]
atomiccupcake: I would love to see an experiment in that. I don't think you'll be able to have it flex to fill vertically because you can't set a parent height that way. Also, with overflow-y, I think there HAS to be an explicit max height.

liwords-ui/src/shared/follower.tsx
atomiccupcake: Keeping it consistent with Blocker (Cesar's code) for now, but you can submit a new pr changing both.

liwords-ui/src/chat/chat.tsx
```diff
@@ -373,6 +386,33 @@ export const Chat = (props: Props) => {
...
+            : acc,
+        undefined
+      );
+      if (
+        lastMessage?.channel !== 'chat.lobby' &&
+        (lastMessage?.timestamp || 0) > lastNotificationTimestamp
+      ) {
+        setNotificationCount((x) => x + 1);
+        setLastNotificationTimestamp((x) =>
+          lastMessage?.timestamp || 0 > x ? lastMessage?.timestamp || x : x
```
andy-k, line 405: if [without parentheses](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Operator_Precedence#table) the ternary's condition is `lastMessage?.timestamp || (0 > x)`, it doesn't look right?

sidenotes:
- `lastMessage?.timestamp` (from golang) and `lastNotificationTimestamp` (from browser's `Date.now()`) may not be in sync -- just a note, maybe not trivial to fix
- `??` conveys the intent slightly better (but it's ok to use `||`)

liwords-ui/src/store/socket_handlers.ts
atomiccupcake: I can't get the typecasting to get past the compiler since PE.toObject has extra fields (deleting, anon). Leaving it. Feel free to patch later?

liwords-ui/src/chat/chat.tsx
atomiccupcake: Ignoring the sidenotes for now. We should address the time issue if we see problems.

liwords-ui/src/chat/players.tsx
atomiccupcake: Nice catch. You get duplicates in club games, since your subscribed to both the club chat and the game or gametv chat. Refactoring.

liwords-ui/src/lobby/seek_form.tsx
```diff
@@ -275,6 +292,15 @@ export const SeekForm = (props: Props) => {
...
   };
 
+  useEffect(() => {
+    if (usernameOptions.length === 0) {
+      const defaultOptions = tournamentID
+        ? presences.map((p) => p.username).filter((u) => u !== username)
+        : Object.values(friends).map((f) => f.username);
+      setUsernameOptions(defaultOptions);
+    }
+  }, [friends, presences, username, usernameOptions, tournamentID]);
```
andy-k, line 302: when the modal to create game is shown (whether Play a computer, Match a friend, or Create a game) the screen froze.
```
Warning: Maximum update depth exceeded. This can happen when a component calls setState inside useEffect, but useEffect either doesn't have a dependency array, or one of the dependencies changes on every render.
    in SeekForm (at gameLists.tsx:228)
```
this is because `defaultOptions` is always a new object, and `setUsernameOptions` re-triggers this effect.

(ps. feel free to merge once it works! thanks for this feature ❤️ )

atomiccupcake: Thanks for all your hard work on this. The defaultOptions fix you mentioned above fixed this.

atomiccupcake: Turns out this wasn't fixed. Infinite loop was caused by setting usernameOptions without a guard in an effect that had usernameOptions as a dependency. Fixed in master.

PR #585, opened 2021-05-04
Removing UN. It's not a country, and we don't have a flag for it.

atomiccupcake:


domino14:
I uploaded one and it's hard right now to edit profiles. I'm going to add an issue to allow mods or admins to edit profiles if needed.

PR #596, opened 2021-05-06
Child Mode

jvc56:


pkg/stores/user/db.go
```diff
@@ -467,8 +470,9 @@ func (s *DBStore) SetPersonalInfo(ctx context.Context, uuid string, email string
 			return result.Error
 		}
 
-		return tx.Model(p).Update(map[string]interface{}{"first_name": firstName,
+		return s.db.Model(p).Update(map[string]interface{}{"first_name": firstName,
```
domino14, line 473: how come not tx?

jvc56: Oops, that was an oversight.

atomiccupcake:
Front end will warn users once a day if they haven't submitted a birthdate. Will hide profile info and full names for children and users of unknown age on the profile. Games will no longer show full names for anyone. (Making a new issue to correct.)

liwords-ui/src/socket/socket.ts
```diff
@@ -107,10 +109,26 @@ export const LiwordsSocket = (props: {
...
+          isChild: decoded.cs,
           path: pathname,
           perms: decoded.perms?.split(','),
         },
       });
+      if (
+        parseInt(decoded.cs) === 2 &&
+        (!localStorage?.getItem('birthdateWarning') ||
+          Date.now() - parseInt(localStorage?.getItem('birthdateWarning')!) >
+            24 * 3600)
```
andy-k, line 121: isn't `Date.now()` in milliseconds?
(also it'd be preferable not to `localStorage.getItem` twice)

liwords-ui/src/socket/socket.ts
```diff
@@ -114,11 +114,12 @@ export const LiwordsSocket = (props: {
...
       });
+      const bdateWarning = localStorage?.getItem('birthdateWarning');
       if (
         parseInt(decoded.cs) === 2 &&
-        (!localStorage?.getItem('birthdateWarning') ||
-          Date.now() - parseInt(localStorage?.getItem('birthdateWarning')!) >
-            24 * 3600)
+        decoded.a &&
+        (!bdateWarning ||
+          Date.now() - parseInt(bdateWarning) > 24 * 3600 * 1000)
```
atomiccupcake, line 122: Nice catch

pkg/profile/profile.go
```diff
@@ -78,7 +78,7 @@ func (ps *ProfileService) GetProfile(ctx context.Context, r *pb.ProfileRequest)
 		FirstName:       user.Profile.FirstName,
```
atomiccupcake, line 78: First and last name should also be suppressed if youth

atomiccupcake (approved):

pkg/profile/profile.go
atomiccupcake: And the "About"

liwords-ui/src/settings/personal_info.tsx
```diff
@@ -298,6 +309,34 @@ export const PersonalInfo = React.memo((props: Props) => {
...
+              <>
+                Date of birth{' '}
+                <span className="notice">(This will not be displayed.)</span>
+              </>
+            }
+            rules={[
+              {
+                required: true,
+                message:
+                  'Your profile information will be private unless you provide a birthdate.',
```
andy-k, line 326: if DOB is mandatory, this message should be reworded - this message seems to say it's fine to have a private bio

PR #601, opened 2021-05-07
follow active games

andy-k:


pkg/stores/user/update_active_game.lua
```diff
@@ -0,0 +1,61 @@
+-- Arguments to this Lua script:
+-- gameuuid expiry p0uuid p1uuid ...
+-- (the script should work with any number of players)
+
+local gameuuid = ARGV[1]
+local expiry = tonumber(ARGV[2]) -- seconds
+
+local ts = tonumber(redis.call("TIME")[1])
```
domino14, line 8: the time should be passed in as an argument to the script to make it deterministic / easier to test

PR #615, opened 2021-05-12
Tournament/Club customization

atomiccupcake:
Backend changes to add color and logo, admin page updates for adding logo, color, disclaimer, board theme, and team theme, show logo and disclaimer on tourney page and game page. Hardcoded settings for the NASSC rooms.


liwords-ui/src/gameroom/table.tsx
```diff
@@ -873,13 +874,21 @@ export const Table = React.memo((props: Props) => {
         : singularCount(n, 'Player', 'Players'),
     [isObserver]
   );
+  const boardTheme =
+    'board--' + tournamentContext.metadata.getBoardStyle() || '';
+  const tileTheme = 'tile--' + tournamentContext.metadata.getTileStyle() || '';
```
domino14, line 879: I think these need to be in parentheses: 

```('tile--' + tournamentContext.metadata.getTileStyle()) || '';``` otherwise the result looks like `tile--` if getTileStyle is empty. same for boardTheme above it.

PR #619, opened 2021-05-13
fix first letter avatar

andy-k:
otherwise when an adult with no avatar image plays, the letter on the avatar in the game room is from the username instead of from the full name


andy-k:
for #596

Issue #609, opened 2021-05-10
Reenable full names in game 


atomiccupcake:
In #596 we disable showing full names in the game room. We need the back end to only provide the full name for known adults, and then revert the front end to show the full name instead of the nickname when available.

PR #620, opened 2021-05-13
hide profile fields

andy-k:
for #596 
- when getting own profile you see everything including BirthDate
- when getting others' profile you don't get BirthDate*
- when getting others' profile you only get FirstName, LastName, FullName, About, AvatarUrl if they are adult
- GetBriefProfile always excludes FullName and AvatarUrl if (at the time they're cached) the subject is not an adult (this means a child can't get their own avatar through brief profiles, because the same cache is shared for every caller)
- no change to the deprecated endpoint
- full name logic change (if first name given but last name missing, no additional space)

\* this breaks frontend, frontend checks should be removed


andy-k:
for #609

domino14 (approved):

PR #634, opened 2021-05-19
Fix Zero Value User Set

jvc56:


pkg/stores/user/db.go
```diff
@@ -178,7 +178,15 @@ func (s *DBStore) Set(ctx context.Context, u *entity.User) error {
 	}
 
 	result := s.db.Model(&User{}).Set("gorm:query_option", "FOR UPDATE").
-		Where("uuid = ?", u.UUID).Update(dbu)
+		Where("uuid = ?", u.UUID).Update(map[string]interface{}{"uuid": dbu.UUID,
```
andy-k, line 181: seems risky to change `.Set` to do this, not sure if all callers expect the new behavior. `GetByEmail` doesn't currently load `Actions` for example.

btw that old `"FOR UPDATE"` isn't relevant here (it'd only be useful for selects in a transaction, not for updates) but probably harmless.

jvc56: Good points, I see if I can do this in a better way.

jvc56: Closing this pull request. Moving solution to the automod branch.

PR #661, opened 2021-05-30
Introduce word wrap to chat window

nibiew:
Stop long words (e.g. links) from extending chat window


andy-k (approved):
👍 now works on safari too

PR #663, opened 2021-05-31
support word smog definitions if backend uses crlf

andy-k:
actually this PR is not necessary, restarting/rebuilding the "app" docker should already resolve it


andy-k:
this isn't it. root cause for the TypeError is the backend CSW19.txt is `"WORD\tDefinition\r\n"` (Windows line ending). only `"WORD\tDefinition\n"` (Unix line ending) is supported. not sure if frontend or backend should work around bad data. it's easy enough to fix the bad data at source:
```
cd ~/git/liwords/data/lexica/words
perl -i -pe 's/\r//g' CSW19.txt
```

Jayem2k20:
Thanks guys! I don't know how those sneaky carriage returns got in there. All sorted now though!

andy-k:
yay!

PR #668, opened 2021-06-01
enhance analyzer

andy-k:
- ability to show equity difference from top move (not enabled, it's always false for now; where/how to place the toggle is tbd)

current:
Screenshot 2021-06-01 at 05 00 00

if enabled: 
Screenshot 2021-06-01 at 04 58 56

- identify actually-played move (styling tbd; for now shown with ">")

Screenshot 2021-06-01 at 11 07 29

- identify invalid plays (styling tbd; for now shown with "\*") (note: play may be invalid even if main word is valid, if a secondary word is invalid, hence do not style this by placing "\*" next to main word) (invalid play would still be in order of equity assuming unchallenged)

Screenshot 2021-06-01 at 11 07 54

- if actually-played move is not in analyzer top few, insert "..." before (and compute the score and equity since it's not given by analyzer)

Screenshot 2021-06-01 at 11 08 11

- works with exch (and pass) too

Screenshot 2021-06-01 at 11 10 00

### DIFF
--- a/liwords-ui/src/gameroom/analyzer.tsx
+++ b/liwords-ui/src/gameroom/analyzer.tsx
@@ -637,11 +637,9 @@ export const Analyzer = React.memo((props: AnalyzerProps) => {
             onClick={() => {
               placeMove(m);
             }}
+            {...((m.chosen ?? false) && { className: 'move-chosen' })}
           >
-            <td className="move-coords">
-              {(m.chosen ?? false) && <React.Fragment>&gt;</React.Fragment>}
-              {m.coordinates}
-            </td>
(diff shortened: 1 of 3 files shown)


domino14:
Andy: design (Conrad) is going to take a pass at the analyzer overall

For now, some feedback from Lola
- we’ll want some different way to show that a play is selected and on the board (but not the `>`), making the background of the row “gray-subtle” color seems okay.
- Putting the … seems redundant. Just adding it to the end and calculating is just fine.
- The toggle for showing equity loss instead of equity score would go in “secret features”. “Show equity loss in analyzer.”

PR #675, opened 2021-06-04
deduplicate phony report

andy-k:
bandaid patch to resolve repeated phony report when reconnecting socket
(view with ?w=1 / ignore whitespace)


domino14 (commented):
This is ok; is it too complex to prevent the repeated XHRs in the first place?

andy-k:
it seems it wasn't doing repeated XHRs, just repeated reporting

PR #676, opened 2021-06-04
reconnect socket on connection timeout

andy-k:


andy-k:
previous attempt was in #288

PR #677, opened 2021-06-04
swap tiles when dragging over a tentative tile on board

andy-k:
this only works on desktop.

on mobile, dragging is handled by something else and i can't figure out what/where.


andy-k:
now works on mobile too

PR #684, opened 2021-06-05
fix mobile tile preview crash

andy-k:
#683 

`Cannot read property 'toLowerCase' of undefined`
in
https://github.com/domino14/liwords/blob/9f315093d8841b369ab3a4b7316507d4b50538d5/liwords-ui/src/utils/cwgame/common.ts#L60-L62
in TilePreview
https://github.com/domino14/liwords/blob/9f315093d8841b369ab3a4b7316507d4b50538d5/liwords-ui/src/gameroom/tile.tsx#L100-L102


andy-k:
why is typescript not catching this?

Screenshot 2021-06-05 at 13 05 40

PR #697, opened 2021-06-10
use valid equity loss only

andy-k:
if "show equity loss" is enabled and the actual played move is a phony, the equity loss was relative to that phony.

this fixes it so that the equity loss is relative to the top valid play instead. (the phony has equity loss of the opposite sign)


liwords-ui/src/gameroom/analyzer.tsx
```diff
@@ -644,6 +644,11 @@ export const Analyzer = React.memo((props: AnalyzerProps) => {
     () => localStorage.getItem('enableShowEquityLoss') === 'true',
     []
   );
+  const equityBase = React.useMemo(
+    () =>
+      showEquityLoss ? moves?.find((x) => x.valid ?? true)?.equity ?? 0 : 0,
```
domino14, line 649: `?`

PR #698, opened 2021-06-10
expose phony in analyzer

andy-k:
need ux input for this.

note that we shouldn't put `*` at the end of the main word. styling will be the same for valid words that make phony hooks.

i propose this:

Screenshot 2021-06-08 at 14 45 48


domino14:
Conrad is redesigning this. My thought is this is probably OK to deploy for now but I'll ask for feedback.

dacronbb:
is this necessary? I'd rather a second line with a description

1) i don't understand the emoji
2) so if the main word is phoney or a hook word is phoney the * shows up in a different place?

andy-k:
it's necessary to display the played phony differently, people are absorbing/learning those phonies when analyzing others' games, they're thinking that the moves are valid because the analyzer suggested them.

the emoji is 🙈 but any other emojis work fine too. alternatively second line might also work (i'm not sure if it'd mess up the alignment), but i think the first line still must look different (otherwise it'd run into the same issue). ideas?

at the moment the library returns valid:true/valid:false, but valid:false doesn't tell if the main word is invalid or if the main word is valid (and the invalidity came from the hooks). so valid:false will be displayed the same way. if it's decided that we need to display these cases differently and/or show the bad formed words, then the library has to be changed too. then we could do "(A)URATiNG*" or "(A)URATiNG* (IQ*)" or "GUARd(A)NT (QU*, NI*)" for example. would that be useful?

what about the * next to the equity? (it means "footnote: if unchallenged")

domino14:
@dacronbb

dacronbb:
i dont think the emoji is quite at the point where people will understand it if they're not competitive players. saying "this play has a phoney" but not identifying the phoney is marginally useful at best. showing the * indication in two separate spots is not helpful for players. I would advice a second line actually identifying the phoney word (which will show up with future designs)

this widget is going to get overhauled soon when it gets prioritized for design, anyway, so i guess do whatever for now.

andy-k:
let's do this?

Screenshot 2021-06-14 at 10 47 27

PR #699, opened 2021-06-10
remove letter offset

andy-k:
need ux approval

currently Q and Z are offset to accommodate 10 pt, this makes reading vertical words slower
Screenshot 2021-06-10 at 13 04 59

this removes the offset and adjusts the letter-spacing
Screenshot 2021-06-10 at 13 04 44

and make it unnecessary to be language-aware (other-language tiles have the 10-pt on different letters)

### DIFF
--- a/liwords-ui/src/gameroom/scss/gameroom.scss
+++ b/liwords-ui/src/gameroom/scss/gameroom.scss
@@ -709,19 +709,6 @@ p.streak-loss {
   text-align: center;
   width: $tile-size-mobile;
   height: $tile-size-mobile;
-  &[data-rune="Q"] {
-    p.rune {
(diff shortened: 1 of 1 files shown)


dacronbb:
are the only changes kerning + removing the offset?

noting that there's a tradeoff in the kerning you've done - it's worse for newer players. That said overall it's an improvement so yeah approved from me

andy-k:
yeah, only the kerning and offset. i compensated for the reduced kerning by adding offset from the right side, that way single-digit values remain at the same place, and the only tiles affected are the Q and Z. i found it interesting that Q (-2px) and Z (-1px) had different offsets although they were both 10pt.

Issue #627, opened 2021-05-14
Add birthday info to sign-up


dacronbb:
[see figma mock](https://www.figma.com/file/XTG9fsH9UO8pSBTHNuUsBD/Woogles-for-Engineering?node-id=4712%3A5625)

andy-k:
@dacronbb for a more inclusive experience please consider
- y-m-d (instead of m-d-y)
- [having Full Name instead of First+Last](https://www.kalzumeus.com/2010/06/17/falsehoods-programmers-believe-about-names/)

Jayem2k20:
It seems counter-intuitive to enter a date as YYYY-MM-DD. I've tweaked the code on the Personal info page to ask for it in the format DD-MM-YYYY (I'm in the UK). It still saves it to the DB as YYYY-MM-DD

andy-k:
i brought it up because _personally_ i find asking for m-d-y _offensive_ to people who aren't using that sequence such as myself. it seems y-m-d would be most acceptable to both m-d-y and d-m-y users.

(the problem with an input textbox asking for dd-mm-yyyy in the uk version is, when someone who's born in the first twelve days of a month keys in mm-dd-yyyy, that would go unnoticed; using three fields avoids this if the month dropdowns show names instead of numbers.)

Jayem2k20:
It does get confusing!

I noticed with the Ant Design Datepicker, it's possible to use localization, 

Would it be a huge task to localize the date setting? Either globally or using the Country setting on an individual basis?

dacronbb:
We ask that front end changes are not made without sign-off from design team (CBB or Lola)!

1) **I'd prefer it be localized if possible**, but I am on board with Andy's suggestion if not. While being _offended_ seems like overkill - I don't mind when Asian airlines ask me for day first, for example. :P

2) The instruction from JD was to include date, yes. People will just lie if they don't know - this is a COPPA thing, not a "we need your birthday to display it" thing.

3) I'm fine w/ First name and Last name as is - each field is optional :P 

4) I would expect the field to allow typing or selection via dropdown. Dunno, 1920 or so seems good :P

domino14:
We looked at a number of different screenshots of different COPPA compliant sign up pages, and they all have MM - DD - YYYY. I think as long as they are three separate boxes, and not one single widget, it is fine. The backend would still save as YYYY-MM-DD. Would localization flip the order of the first two boxes?

e.g.

[...]

PR #710, opened 2021-06-15
birthday sign up

andy-k:
#627 

in-scope but not included (yet?)
- [ ] PP and TOS links, for now they're placeholders
- [x] "log in here" link, for now it links to lobby since it doesn't have its own page
- [x] birth date as 3 dropdowns (maybe can't show the valid/invalid feedback), for now it's a yyyy-mm-dd text box (not a date picker)

out-of-scope for this PR
- country dropdown (it's copied from the one in settings page) still has Unspecified and can't show flags on Windows
- no validation for birth date range (because there's none in settings page either)


domino14:
Thanks so much! I think this is fine to merge and we'll update the TOS link when it's ready (@magratheazaphod )

PR #711, opened 2021-06-16
Remove space from scrubbed username and scrub userid

jvc56:


pkg/mod/mod.go
```diff
@@ -20,7 +20,7 @@ import (
 
 // Spaces are never allowed in usernames,
 // so this string will never be a valid username.
-var CensoredUsername = "Unknown Woogler"
+var CensoredUsername = "Unwoogler"
```
andy-k, line 23: - comment mismatch now
https://github.com/domino14/liwords/blob/9988ce136a9f488397fc29256d8800fbe21cffd4/pkg/mod/mod.go#L21-L23
- need to patch this
https://github.com/domino14/liwords/blob/e2b70b0cef09d3ea8b92dec2ca12e71f1a4f7b68/pkg/registration/register.go#L27-L30
- Unwoogler now shows up in rematch streak, are we now changing all "Unknown Woogler"?
- full name, country, (avatar?) still show up; turns out the `GAME_HISTORY_REFRESHER` in socket still has the real `userId`
- in game rooms, avatar letter now shows as '?' on player card, but as 'U' on moves list (seems PlayerAvatar is getting called with undefineds)

pkg/mod/mod.go
```diff
@@ -20,7 +20,7 @@ import (
...
 var CensoredAvatarUrl = "https://woogles-prod-assets.s3.amazonaws.com/unknown-woogler.png"
 var CensoredAboutText = "This account does not exist."
```
andy-k, line 25: (also these don't seem to work now?)

pkg/mod/mod.go
jvc56: It was decided that all instances of "Unknown Woogler" can be changed to "Unwoogler" to avoid whitespace problems. So it looks like there is still the GAME_HISTORY_REFRESHER socket events that need to be sanitized. Where is the call to PlayerAvatar happening?

pkg/mod/mod.go
jvc56: What is the problem. Is content not being scurbbed?

domino14 (approved):

pkg/registration/register.go
```diff
@@ -25,7 +26,7 @@ func RegisterUser(ctx context.Context, username string, password string, email s
 		return errors.New("username can only contain letters, digits, period, hyphen or underscore")
 	}
 	// Should we have other unacceptable usernames?
-	if strings.ToLower(username) == "anonymous" {
+	if strings.ToLower(username) == "anonymous" || username == utilities.CensoredUsername {
```
andy-k, line 29: need to disallow registering with alternate capitalization
```suggestion
	if strings.EqualFold(username, "anonymous") || strings.EqualFold(username, utilities.CensoredUsername) {
```

andy-k (approved):

PR #715, opened 2021-06-19
mutual abort touch up

andy-k:
this may help clear half of the remaining tasks, check each commit to find out which. probably won't do the rest. check the added todo comments.


liwords-ui/src/gameroom/game_controls.tsx
```diff
@@ -77,7 +78,7 @@ const ExamineGameControls = React.memo((props: { lexicon: string }) => {
 });
 
 type OptionsMenuProps = {
-  handleOptionsClick: (e: any /* GOD what is the actual type here */) => void;
+  handleOptionsClick: (e: MenuInfo) => void;
```
domino14, line 81: 😅

Issue #716, opened 2021-06-19
add Match User context menu when clicking on a player 


domino14:
it should do the right thing if we're in clubhouse mode as well

PR #717, opened 2021-06-19
add match user context menu

andy-k:
#716


domino14 (changes requested):
The only place I'd like to remove this from is from inside a paired tournament room (i.e. not clubhouse mode). Otherwise I think there will be confusion - all tournament pairings in these rooms should be automatic; the receiver would hear a match sound but not see it on their screen, etc.

PR #718, opened 2021-06-19
Fix permabanned user game history bug

jvc56:


domino14:
merged #714 which solves same issue 😝

Issue #725, opened 2021-06-24
Analyzer card messed up


Jayem2k20:
Not sure if it's something I've done but, If you add a Disclaimer to a Tournament, and then Examine a game played from that tournament, the layout of the Analyzer card gets messed up.

![AnalyzerError](https://user-images.githubusercontent.com/83062931/123232007-a4512700-d4d0-11eb-8026-be6ee6d783a4.png)

Edit: It only seems to be an issue when in "three column" mode

![AnalyzerError1](https://user-images.githubusercontent.com/83062931/123234281-baf87d80-d4d2-11eb-9953-f5194f07c949.png)

![Analyzer1](https://user-images.githubusercontent.com/83062931/123234341-c51a7c00-d4d2-11eb-8c22-17e039354346.png)

![Analyzer2](https://user-images.githubusercontent.com/83062931/123234371-cc418a00-d4d2-11eb-9a40-9943e98cff87.png)

PR #726, opened 2021-06-24
hide disclaimer when examining

andy-k:
simple patch for #725 by always hiding the disclaimer when examining (regardless of screen size)


Jayem2k20:
That works! Thanks Andy

domino14 (approved):

PR #757, opened 2021-07-06
Terms of Service

atomiccupcake:
Add terms of service page. It's a dropdown on the "About Us" in the menu. It does not matter what the initial version date is set to as everyone who registered before then will get it, but should be something reasonably close to deploy date.

### DIFF
--- a/liwords-ui/src/App.tsx
+++ b/liwords-ui/src/App.tsx
@@ -220,9 +220,9 @@ const App = React.memo(() => {
 
   useEffect(() => {
     const notifyInfo = getTermsNotify();
-    if (!notifyInfo?.notified) {
+    if (loggedIn && !notifyInfo?.notified) {
       notification.info({
-        message: `Our terms of service were updated on ${notifyInfo.changed}.`,
+        message: `Our terms of service were updated on ${notifyInfo.changed}. Click to read them.`,
         duration: 0,
         key: 'terms',
         onClick: () => {
@@ -234,7 +234,7 @@ const App = React.memo(() => {
         },
       });
     }
-  }, [history]);
+  }, [history, loggedIn]);
 
   const sendChat = useCallback(
     (msg: string, chan: string) => {
--- a/liwords-ui/src/App.tsx
+++ b/liwords-ui/src/App.tsx
@@ -1,9 +1,16 @@
 import React, { useCallback, useEffect, useRef } from 'react';
-import { Route, Switch, useLocation, Redirect } from 'react-router-dom';
+import {
+  Route,
+  Switch,
+  useLocation,
+  Redirect,
+  useHistory,
+} from 'react-router-dom';
 import { useMountedState } from './utils/mounted';
 import './App.scss';
 import axios from 'axios';
 import 'antd/dist/antd.css';
+import { notification } from 'antd';
 
 import { Table as GameTable } from './gameroom/table';
 import TileImages from './gameroom/tile_images';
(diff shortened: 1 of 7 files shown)


liwords-ui/src/App.tsx
```diff
@@ -204,6 +218,24 @@ const App = React.memo(() => {
     }
   }, [getFriends, pendingFriendsRefresh]);
 
+  useEffect(() => {
+    const notifyInfo = getTermsNotify();
+    if (loggedIn && !notifyInfo?.notified) {
+      notification.info({
+        message: `Our terms of service were updated on ${notifyInfo.changed}. Click to read them.`,
+        duration: 0,
```
domino14, line 226: What do we think about making this fade away eventually, after 5-10 seconds? I worry that a persistent notification might distract people when they join a game?

atomiccupcake: We should ask Conrad. More like 20 seems better to me for this. It could also just close if they start a game or whenever we choose.
Tangentially, are you still not seeing the X to close those notifications? Wonder if you have user CSS on from the community styles or something.

dacronbb: if we're not forcing people to accept anything, why don't we just send an email to all users instead of an in-app notification?

That seems less distracting, and the pattern companies usually use AFAIK.

liwords-ui/src/about/termsOfService.tsx
```diff
@@ -0,0 +1,444 @@
...
+export const setTermsNotify = () => {
+  localStorage?.setItem('termsDate', Date.parse(TERMS_VERSION_DATE).toString());
+};
+
+export const getTermsNotify = () => {
+  const releaseDate = Date.parse(TERMS_VERSION_DATE);
+  const termsDate = parseInt(localStorage?.getItem('termsDate') || '0', 10);
+  return {
+    notified: releaseDate === termsDate,
+    changed: moment(TERMS_VERSION_DATE).format('MMMM Do, YYYY'),
```
andy-k, line 19: - US date format only
- just return `changed: TERMS_VERSION_DATE,` and let the caller format only when needed
- similarly the localStorage can store the TERMS_VERSION_DATE as-is (instead of epoch millis as string) to skip Date.parse/toString/parseInt

liwords-ui/src/topbar/topbar.tsx
```diff
@@ -19,6 +19,20 @@ import { isClubType } from '../store/constants';
...
+  const aboutMenu = (
+    <ul>
+      <li>
+        <Link className="plain" to="/team">
+          Meet the team
+        </Link>
+      </li>
+      <li>
+        <Link className="plain" to="/terms">
+          Terms of service
```
andy-k, line 31: special exception for not using sentence case since Terms of Service is a name
```suggestion
          Terms of Service
```

liwords-ui/src/about/termsOfService.tsx
```diff
@@ -0,0 +1,444 @@
...
+            state or federal courts located in Essex County, New Jersey, and you
+            hereby consent and submit to the venue and personal jurisdiction of
+            such courts for the purposes of such action.
+          </p>
+          <p>
+            In the event any provision of these terms is held to be invalid or
+            unenforceable, then that provision will be limited to the minimum
+            enforcement possible, and the remaining terms held with full effect.
+          </p>
+          <p>These terms are an agreement between you and Woogles.</p>
```
andy-k, line 439: "These terms are" (plural) and "an agreement" (singular) seem to disagree, not sure if it's correct or how to rephrase

liwords-ui/src/about/termsOfService.tsx
domino14: US date format is entirely within local storage right? (in any case, this code has been backed out)

domino14:
@andy-k 

I did not implement the Oxford comma or any other wording changes. As this document has been actually looked through by a lawyer I don't think we should change them. @magratheazaphod I'll defer to you on these.

domino14:
Thanks for looking at this thoroughly!

PR #766, opened 2021-07-19
Fix the things

jvc56:


pkg/tournament/tournament.go
```diff
@@ -896,14 +896,16 @@ func DivisionChannelName(tid, division string) string {
 }
 
 func StartRoundCountdown(ctx context.Context, ts TournamentStore, id string,
-	division string, round int, save, send bool) error {
+	division string, round int, save, send, lock bool) error {
 	t, err := ts.Get(ctx, id)
```
andy-k, line 900: there still is no guarantee that this t is the same as the one StartAllRoundCountdowns will be saving

jvc56: I think this will be OBE by the time it becomes a problem.

PR #770, opened 2021-07-21
prevent mutation on error

domino14:


domino14:
need to actually test this locally

PR #776, opened 2021-07-24
Refactor Tournament Start

jvc56:


pkg/tournament/tournament.go
```diff
@@ -879,7 +943,7 @@ func StartAllRoundCountdowns(ctx context.Context, ts TournamentStore, id string,
 
 	for division := range t.Divisions {
 		t.IsStarted = true
-		err := StartRoundCountdown(ctx, ts, id, division, round, false, true, false)
+		err = startRoundAfterLockingTournament(ts, t, ctx, division, round, false)
```
andy-k, line 946: the above change of send=true to send=false seems unintended
```suggestion
		err = startRoundAfterLockingTournament(ts, t, ctx, division, round, true)
```

jvc56: I was the big confused. After talking to @domino14 it seems that this send boolean is always set to true currently. It probably had a purpose at some point that was lost in a refactor :/.

andy-k: `startRoundAfterLockingTournament` actually has three parts:
1. pre-check
1. call StartRound
1. send event

the "pre-check" part is quite thorough, but `StartAllRoundCountdowns` has its own pre-check, to try to avoid having later divisions error out after starting earlier divisions. they might do (almost) the same thing for now, but if the pre-check is factored out such that `StartAllRoundCountdowns` can call it too (for a minor performance penalty), this would ensure that they are really checking the same things. maybe that way the `t.Started = true` can be moved to just before `ts.Set`.

the "call `StartRound`" part is effectively a one-liner.

the "send event" part (possibly slow especially if the buffered channel is full) should probably be split out, to be called (by both `StartAllRoundCountdowns` and `StartRoundCountdown`) only after `ts.Set` succeeds. if `ts.Set` fails, the in-memory cache is corrupted, but at least the frontend would reflect what's in db. (this has its own problems though, not sure which way makes more sense.)

(the `send` parameter is still not needed)

jvc56: Hmm, I'm not sure I understand what the actionable items are here.

pkg/tournament/tournament.go
```diff
@@ -852,126 +917,77 @@ func StartAllRoundCountdowns(ctx context.Context, ts TournamentStore, id string,
...
+	if err != nil {
+		return err
 	}
 
-	/*	We will deal with auto start at a later date
-		if manual && divisionObject.Controls.AutoStart && divisionObject.DivisionManager.IsStarted() {
-			return fmt.Errorf("division %s has autostart enabled and cannot be manually started", division)
-		}*/
-
-	if !t.IsStarted {
```
andy-k, line 929: it seems the new `StartRoundCountdown` no longer requires `t.IsStarted`. is that intended?

the rest of the PR seems ok.

jvc56: Yes, that was intentional. I realized that if we keep that check and some of the rounds on SARC fail, then we have no way to individually start the rounds that had failed, since t.Started was never set to true. In this way we can regard SARC is a batch call identical to calling SRC individually.

andy-k: it almost is, but SARC also sets t.IsStarted. if you call each SRC individually then t.IsStarted will never be set and that'd break other stuffs (and you can no longer call SARC because they'd be at different rounds). we don't have a "just set t.IsStarted" as a separate step. we probably should, and then both SARC and SRC can check that this has been done. not sure.

jvc56: Ah, good point. I think the most straightforward way is to just have both SRC and SARC set t.Started = true on success.

andy-k (approved):

PR #780, opened 2021-08-03
Automod Notifications

jvc56:


pkg/mod/mod.go
```diff
@@ -322,66 +315,7 @@ func applyAction(ctx context.Context, us user.Store, cs user.ChatStore,
...
-					} else if resp.StatusCode != 204 { // No Content
-						// We do not expect any other response
-						log.Err(err).Str("status", resp.Status).Msg("mod-action-discord-notification-post-bad-response")
-					}
-				}
-			}
-		}
-	}
+	notify(ctx, us, user, action)
 	return us.Set(ctx, user)
```
andy-k, line 319: not sure if `notify` should be before `us.Set` succeeds

pkg/mod/automod.go
```diff
@@ -175,11 +175,13 @@ func updateNotoriety(ctx context.Context, us user.Store, ns NotorietyStore, user
...
-			err = setCurrentAction(user, &ms.ModAction{UserId: user.UUID,
+			action := &ms.ModAction{UserId: user.UUID,
 				Type:          ms.ModActionType_SUSPEND_GAMES,
 				StartTime:     ptypes.TimestampNow(),
 				ApplierUserId: AutomodUserId,
-				Duration:      int32(DurationMultiplier * (newNotoriety - NotorietyThreshold))})
+				Duration:      int32(DurationMultiplier * (newNotoriety - NotorietyThreshold))}
+			err = setCurrentAction(user, action)
+			notify(ctx, us, user, action)
 			if err != nil {
```
andy-k, line 185: not sure if `notify` should be before `setCurrentAction` and `us.Set` succeed

pkg/mod/mod.go
jvc56: `notify` after `us.Set` makes more sense. I'll update it.

pkg/mod/automod.go
jvc56: `notify` after `us.Set` makes more sense. I'll update it.

andy-k (approved):

Issue #787, opened 2021-08-04
Add random tile order option


atomiccupcake:
Add "random" as a rack tile order preference.

PR #800, opened 2021-08-09
add auto shuffle option

andy-k:
for #787 - please adjust styling etc.

the "default tile order" preference must be deterministic (it's used for "reset rack", scorecard, analyzer, notepad's plus button, and maybe others). so instead of adding a "random" option there, this PR adds a new option to automatically shuffle the rack at (the start of) every turn. "reset rack" will still re-sort (alphabetize or whatever the preferred order is).

because we keep a backup of one turn (to reinstate premove on successful challenge, see #759), even if toggling this option directly from the board were possible, it would only take effect two turns later. conversely, because we only keep a backup of one turn, with auto shuffle enabled, navigating from turn A to turn B to turn C and back to turn A will not recover rack A's ordering, it will be reshuffled anew.

Screenshot 2021-08-09 at 18 50 12

### DIFF
--- a/liwords-ui/src/settings/preferences.tsx
+++ b/liwords-ui/src/settings/preferences.tsx
@@ -1,7 +1,12 @@
 import React, { useCallback } from 'react';
 import { useMountedState } from '../utils/mounted';
 import { Col, Row, Select, Switch } from 'antd';
-import { preferredSortOrder, setPreferredSortOrder } from '../store/constants';
+import {
+  preferredSortOrder,
+  setPreferredSortOrder,
+  setSharedEnableAutoShuffle,
+  sharedEnableAutoShuffle,
+} from '../store/constants';
 import '../gameroom/scss/gameroom.scss';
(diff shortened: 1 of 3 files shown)


domino14:
Thank you for implementing this. Could we make this part of the default tile order dropdown instead, inside the games?

liwords-ui/src/settings/preferences.tsx
```diff
@@ -185,11 +193,90 @@ export const Preferences = React.memo((props: Props) => {
...
+      name: string,
+      value: string,
+      autoShuffle: boolean
+    ) => {
+      // Design only wants "Random" for "Alphabetical".
+      if (
+        !(
+          (tileOrder === value && autoShuffle === localEnableAutoShuffle) ||
+          name === 'Alphabetical' ||
+          !autoShuffle
```
domino14, line 233: I'm not sure what this does - does it only add the `Random` option in some cases?

Issue #786, opened 2021-08-04
Highlight bingos in the scorecard


atomiccupcake:
We would like to have bingos displayed differently in the scorecard than other plays

PR #802, opened 2021-08-09
highlight bingos in the scorecard

andy-k:
just a placeholder for #786, feel free to take it from here. (the css is commented out)

open questions
- how to style them
- what about phonies
- what about valid bingos forming phony hooks
- what if challenged successfully
- what if challenged unsuccessfully for each rule
- regardless how the bingos are styled in scorecard, how to provide a separate copyable list


atomiccupcake:
image

PR #808, opened 2021-08-12
Seek Upgrades

jvc56:
This is untested draft work. I'm just push to make sure this is the right direction.


pkg/bus/meseeks.go
```diff
@@ -44,6 +55,28 @@ func (b *Bus) seekRequest(ctx context.Context, auth, userID, connID string,
 		return err
 	}
 
+	exists, err := b.soughtGameStore.ExistsForUser(ctx, gameRequest.RequestId)
+	if err != nil {
+		return err
+	}
+
+	if exists {
+		err = b.updateSeekRequest(ctx, auth, userID, connID, req)
```
jvc56, line 64: Need to check who is attempting to modify this seek and maybe reject it.

pkg/bus/meseeks.go
```diff
@@ -86,6 +122,49 @@ func (b *Bus) seekRequest(ctx context.Context, auth, userID, connID string,
...
+	// If we are here the seek exists and the game request is not nil
+	if newReq.GameRequest == nil {
+		return errors.New("nil game request for seek to update")
+	}
+	reqId := newReq.GameRequest.RequestId
+	sg, err := b.soughtGameStore.Get(ctx, reqId)
+	if err != nil {
+		return err
+	}
+
```
jvc56, line 137: Need to check who is attempting to modify this seek and maybe reject it.

pkg/bus/meseeks.go
```diff
@@ -86,6 +122,49 @@ func (b *Bus) seekRequest(ctx context.Context, auth, userID, connID string,
...
+	if userID == sg.Seeker() {
+		sg.SeekRequest.UserState = newReq.UserState
+		sg.SeekRequest.BootedReceivers = newReq.BootedReceivers
+	} else {
```
jvc56, line 141: Make this else if and then error if user is not UserID or Receiver

pkg/bus/meseeks.go
```diff
@@ -86,6 +122,49 @@ func (b *Bus) seekRequest(ctx context.Context, auth, userID, connID string,
...
+	}
+
+	// If both players are ready, start the game
+	// If the receiver is absent, send a new seek to everyone
+	// Otherwise, only send it to the two players
+	if sg.SeekRequest.ReceiverState == pb.SeekState_READY &&
+		sg.SeekRequest.UserState == pb.SeekState_READY {
+		return b.gameAccepted(ctx, &realtime.SoughtGameProcessEvent{RequestId: reqId}, userID, connID)
+	} else if sg.SeekRequest.ReceiverState == pb.SeekState_ABSENT {
+		// Send to everyone, the seek is open
```
jvc56, line 152: There is code that deletes the seeks of someone who leaves. In this case we just need to modify this seek state.

pkg/bus/meseeks.go
```diff
@@ -86,6 +122,49 @@ func (b *Bus) seekRequest(ctx context.Context, auth, userID, connID string,
...
+		return b.newSeekRequest(ctx, auth, userID, connID, newReq)
```
jvc56, line 153: Need to resend the existing seek, not send a new one.

pkg/bus/meseeks.go
```diff
@@ -86,6 +122,49 @@ func (b *Bus) seekRequest(ctx context.Context, auth, userID, connID string,
...
+	} else {
+		// Update the current seek request
```
jvc56, line 155: Two cases for sending the seek:

- Send to the whole world (there is an ABSENT state or receiver goes from absent to not absent)
- Only send to players inwolwed (otherwise)

pkg/bus/meseeks.go
```diff
@@ -86,6 +122,49 @@ func (b *Bus) seekRequest(ctx context.Context, auth, userID, connID string,
 	return nil
 }
 
+func (b *Bus) updateSeekRequest(ctx context.Context, auth, userID, connID string,
```
jvc56, line 125: Rename UserID to Seeker

api/proto/realtime/realtime.proto
```diff
@@ -172,13 +172,23 @@ message PresenceEntry {
...
+  READY = 2;
+}
+
 message SeekRequest {
   GameRequest game_request = 1;
   MatchUser user = 2;
   int32 minimum_rating = 3;
   int32 maximum_rating = 4;
   // connection_id is the websocket ID via which this game was requested.
   string connection_id = 5;
```
jvc56, line 187: Add field for the connection id of the receiver.

pkg/bus/meseeks.go
```diff
@@ -86,6 +122,49 @@ func (b *Bus) seekRequest(ctx context.Context, auth, userID, connID string,
...
+	} else if sg.SeekRequest.ReceiverState == pb.SeekState_ABSENT {
+		// Send to everyone, the seek is open
+		return b.newSeekRequest(ctx, auth, userID, connID, newReq)
+	} else {
+		// Update the current seek request
+		evt := entity.WrapEvent(sg.MatchRequest, pb.MessageType_MATCH_REQUEST)
+		log.Debug().Interface("evt", evt).Interface("req", newReq).Interface("receiver", sg.MatchRequest.ReceivingUser).
+			Str("sender", userID).Msg("publishing changed state match request to user and receiver")
+		b.pubToUser(userID, evt, "")
+		b.pubToConnectionID(connID, userID, evt)
```
jvc56, line 160: Need to publish to receiver using new field.

pkg/bus/meseeks.go
```diff
@@ -100,6 +179,10 @@ func (b *Bus) matchRequest(ctx context.Context, auth, userID, connID string,
 		return err
 	}
 
+	if len(req.BootedReceivers) > BootedReceiversMax {
```
jvc56, line 182: Do matches have a boot option?

pkg/bus/meseeks.go
```diff
@@ -193,21 +276,79 @@ func (b *Bus) matchRequest(ctx context.Context, auth, userID, connID string,
...
+	} else if sg.MatchRequest.ReceiverState == pb.SeekState_ABSENT {
+		evt := entity.WrapEvent(sg.MatchRequest, pb.MessageType_MATCH_REQUEST)
+		log.Debug().Interface("evt", evt).Interface("req", req).Str("sender", userID).Msg("receiver is absent")
+		b.pubToConnectionID(connID, userID, evt)
+	} else {
+		evt := entity.WrapEvent(sg.MatchRequest, pb.MessageType_MATCH_REQUEST)
+		log.Debug().Interface("evt", evt).Interface("req", req).Interface("receiver", sg.MatchRequest.ReceivingUser).
+			Str("sender", userID).Msg("publishing changed state match request to user and receiver")
+		b.pubToUser(req.ReceivingUser.UserId, evt, "")
+		b.pubToConnectionID(connID, userID, evt)
```
jvc56, line 338: Need receiver connection ID to publish to only the seeker and the receiver.

pkg/bus/meseeks.go
```diff
@@ -193,21 +276,79 @@ func (b *Bus) matchRequest(ctx context.Context, auth, userID, connID string,
...
+	b.pubToUser(receiverID, evt, "")
 	// Publish it to the requester as well. This is so they can see it on
 	// their own screen and cancel it if they wish.
-	b.pubToConnectionID(connID, reqUser.UserId, evt)
+	b.pubToConnectionID(connID, userID, evt)
 
 	return nil
 }
 
+func (b *Bus) updateMatchRequest(ctx context.Context, auth string, userID string, connID string, req *pb.MatchRequest) error {
```
jvc56, line 308: This now might be called by the seeker or the receiver. Must make changes accordingly.

atomiccupcake:
Did you want to convert this to a pull request instead of a draft so Cesar can look and merge?

jvc56:
Sure, but this has only been tested on my local machine. Someone else should probably test it too, or write formal test cases.

jvc56:
If your changes will involve implementing Conrad's seek designs, you can work off this branch as well.

pkg/stores/soughtgame/db.go
```diff
@@ -239,38 +240,19 @@ func (s *DBStore) UpdateForReceiverConnID(ctx context.Context, connID string) (*
...
 	var err error
 	ctxDB := s.db.WithContext(ctx)
-	query := ctxDB.Table("soughtgames").
-		Where("receiver = ? AND receiver_is_permanent IS TRUE", receiverID)
+	var queryString string
 	if tourneyID != "" {
-		query = query.Where("request->>'tournament_id' = ?", tourneyID)
+		queryString = fmt.Sprintf("receiver = '%s' AND request->>'tournament_id' = '%s'", receiverID, tourneyID)
+	} else {
+		queryString = fmt.Sprintf("receiver = '%s' OR receiver = ''", receiverID)
```
domino14, line 252: this is definitely prone to potential SQL injections. I think in our case it's not possible for users to change the receiverID and tourneyID parameters to anything bad, but I'm not 100% sure. 

You can replace the `'%s'` with `?` and then have an array like:

`params := []string{receiverID, tourneyID}`

or 

`params := []string{receiverID}` 

then call it like `Where(queryString, params...)` in line 255 below

domino14 (changes requested):

PR #841, opened 2021-09-12
Timeout Tournament Spread

jvc56:


pkg/tournament/tournament_classic_test.go
```diff
@@ -1330,6 +1330,66 @@ func TestClassicDivisionSwissRemoved(t *testing.T) {
 	}
 }
 
+func TestClassicDivisionTimeoutSpread(t *testing.T) {
+	// This test is used to ensure that round robin
+	// pairings work correctly
```
domino14, line 1335: comment here doesn't match what the function does

domino14 (changes requested):
just a tiny change with a comment

PR #842, opened 2021-09-13
Void Tournament Result

jvc56:


pkg/tournament/tournament_classic_test.go
```diff
@@ -1330,6 +1330,77 @@ func TestClassicDivisionSwissRemoved(t *testing.T) {
 	}
 }
 
+func TestClassicDivisionVoidResult(t *testing.T) {
+	// This test is used to ensure that round robin
+	// pairings work correctly
```
domino14, line 1335: comment here too :)

domino14 (changes requested):
comment change

PR #846, opened 2021-09-29
Woogles Errors

jvc56:


domino14 (approved):
this seems like a good start.

atomiccupcake (approved):

pkg/tournament/tournament.go
```diff
@@ -140,14 +140,14 @@ func SetTournamentMetadata(ctx context.Context, ts TournamentStore, meta *pb.Tou
...
-		return fmt.Errorf("tournament %s has finished", meta.Id)
+		return entity.NewWooglesError(realtime.WooglesError_TOURNAMENT_FINISHED, []string{t.Name, ""})
 	}
 
 	t.Lock()
 	defer t.Unlock()
 	name := strings.TrimSpace(meta.Name)
 	if name == "" {
-		return errors.New("name cannot be blank")
+		return entity.NewWooglesError(realtime.WooglesError_TOURNAMENT_EMPTY_NAME, []string{t.Name, ""})
```
domino14, line 150: how come this and line 143 have an extra empty string?

jvc56: good question

domino14:
could u rebase it?

domino14:
@jvc56 i ran into a bug while testing this. Apparently, you can submit a result for a tournament as follows:

- p1 is playing p2
- submit result: p1 beat p1 350-300 (does not get rejected)
- shows up as p1 beating p2 350-300

domino14:
Another bug:

- Four players in a division, Fontes pairings
- Remove one player after starting round 1
- PairRound fails so the player is not removed in the front end
- Upon refreshing, re-add the player we removed
- Error when adding player to "registrants" table (probably because we didn't remove player from registrants since step 3 failed, and it wasn't a transaction)

domino14:
close #888

Issue #877, opened 2021-11-30
app should not disconnect and reconnect to socket when switching to a game


domino14:
instead we should open/close subscriptions to different channels within the socket server.

```
the socket isn't part of the store, if you really want it to not be reset then put it outside
but then you'll have to house the login state elsewhere
the socket needs things like the sendMessage handler, which depend on the chat store, too
dk, can't remember the details
in App.tsx we have <LiwordsSocket .../>, it will connect on mount and disconnect on dismount
in index.tsx: <BrowserRouter><Store><BriefProfiles><App />


if you do <BrowserRouter><LiwordsSocket ... /><Store><BriefProfiles><App /> and App doesn't have socket, then LiwordsSocket will not be reset on resetStore

but the token shouldn't be part of the query params
neither should the path

just connect anonymously, and send a command to turn that anon conn to auth conn

```

domino14:
We get a number of cancelled games every day because somehow a user with a bad connection (or a faulty browser?) doesn't reconnect to the socket when switching to the game page. It looks like a Woogles glitch.

domino14:
big CPU spikes during wespac when observers would move from one game to its rematch. This would help.

PR #898, opened 2022-01-22
reset store without reconnecting socket

andy-k:
partial attempt at #877.

`liwords-socket` needs the path to know which realms to subscribe to. if the path changes, the socket must be killed and reconnected.

on closer inspection, it seems only paths starting with `/game/`, ` /tournament/`, or `/club/` need to be preserved.

this PR
- moves `LiwordsSocket` to between `LoginStateStore` and `RestOfStore`.
- changes the reset whole store on path change to reset only `RestOfStore`, so that in itself this does not cause socket reconnection.
- removes `path` from `LoginStateStore`. it's only used in `TournamentRoom` and is currently always equal to the path returned by `react-router-dom` anyway. also removes `isChild`, which is unused. (there is a [no-op cast in `store/login_state.ts` that seems to not prevent injection of such additional keys during the object spread](https://github.com/domino14/liwords/blob/212d247da63d22a98cee9ae9918242cbb7a375dc/liwords-ui/src/store/login_state.ts#L27-L31).)
- canonicalizes, for the purpose of socket realm subscription, all paths not starting with `/game/`, ` /tournament/`, or `/club/` to `/`. so, there's no reconnection when people go from `/` to `/profile` to `/settings` to `/team` or something like that. (of course it's not very useful if the main load is from rematch redirects.)
- late-binds the socket handler to accommodate the above change. while the socket itself only depends on the login state and canonicalized path, the handler called when a socket message arrives processes that message on the rest of store.
- does nothing about auto socket reconnection, it seems there's already some logic there.

expect even more bugs. test thoroughly and be ready to revert.

obviously, this PR introduces a new bug: socket messages received by the old handler will be processed by the old rest of store, and socket messages received by the new handler will be processed by the new rest of store after reset. (there's also a brief period of time when neither handler is installed, during which received messages are lost permanently.) as long as "reset rest of store" is still a thing, this bug cannot be fixed.

### DIFF
--- a/liwords-ui/src/App.tsx
+++ b/liwords-ui/src/App.tsx
@@ -1,6 +1,5 @@
 import React, { useCallback, useEffect, useRef } from 'react';
 import { Route, Switch, useLocation, Redirect } from 'react-router-dom';
-import { useMountedState } from './utils/mounted';
 import './App.scss';
 import axios from 'axios';
 import 'antd/dist/antd.css';
@@ -17,7 +16,7 @@ import {
   useFriendsStoreContext,
 } from './store/store';
 
-import { LiwordsSocket } from './socket/socket';
+import { useLiwordsSocketContext } from './socket/socket';
 import { Team } from './about/team';
 import { Register } from './lobby/register';
 import { UserProfile } from './profile/profile';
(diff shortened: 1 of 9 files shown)


domino14:
Please rebase when you get the chance; I believe both sets of imports are required now? (in the file with conflicts)

PR #902, opened 2022-01-26
Log Authenticated Tournament Director Requests

jvc56:


pkg/tournament/service.go
```diff
@@ -463,7 +463,7 @@ func directorOrAdmin(ctx context.Context, ts *TournamentService) (*entity.User,
 	return user, nil
 }
 
-func authenticateDirector(ctx context.Context, ts *TournamentService, id string, authenticateExecutive bool) error {
+func authenticateDirector(ctx context.Context, ts *TournamentService, id string, authenticateExecutive bool, reqString string) error {
```
domino14, line 466: let's pass in the req as `proto.Message` and then just do log.Info().Interface("req", req), i think this is a little cleaner?

domino14:
Please rebase when you get the chance

domino14 (approved):

PR #905, opened 2022-01-28
game service skeleton

domino14:


domino14:
When migrating old GameHistory:

- make sure to convert to JSON
- convert to use player indexes like in https://github.com/domino14/macondo/pull/160
- make sure that the Lexicon field is filled in (if not, fill it in from the game's GameRequest column)
- of course migrate GameRequest to JSONB

PR #943, opened 2022-03-12
Puzzles

jvc56:
### Backend implementation of puzzles.

- There is a stores/common package that I created in anticipation of future changes. Some of the functions take a database connection and/or a transaction. I'm not sure what is best, there could probably be improvements in this area.
- Using Sprintf for database strings is not ideal. Look at puzzles/db.go for a more detailed comment, but basically, I could not make the interpolation work in this case for this driver. This can probably be improved.
- authorId hack in db.go CreatePuzzle could probably be improved.
- The puzzle store implements 'GetUserRating', even though it really shouldn't. This is because I did not want to add more tech debt by passing through another store when we plan to pay off this tech debt later. This should be resolved once we eliminate the distinct stores and use a single database connection instead.
- In future work, I plan on consolidating the current utilities package into the common package, as well as consolidating other common functions and testing variables.


pkg/bus/gameplay.go
```diff
@@ -430,10 +434,14 @@ func (b *Bus) sendGameRefresher(ctx context.Context, gameID, connID, userID stri
...
 	defer entGame.RUnlock()
 	var evt *entity.EventWrapper
+	log.Info().Str("gameid", entGame.History().Uid).Msg("sent-refresher")
+
 	if !entGame.Started && entGame.GameEndReason == pb.GameEndReason_NONE {
+		log.Info().Str("gameid", entGame.History().Uid).Msg("sent-refresher-unstarted-game")
 		evt = entity.WrapEvent(&pb.ServerMessage{Message: "Game is starting soon!"},
 			pb.MessageType_SERVER_MESSAGE)
 	} else {
+		log.Info().Str("gameid", entGame.History().Uid).Msg("sent-refresher-for-started-game")
```
domino14, line 444: let's make these three Debug()

api/proto/ipc/omgwords.proto
```diff
@@ -192,6 +192,8 @@ message GameInfoResponse {
   int32 tournament_round = 21;
   // a game index within a round.
   int32 tournament_game_index = 22;
+  // game type, for example native, annotated, bot vs bot, etc.
+  int32 type = 23;
```
domino14, line 196: this should be an enum

pkg/stores/game/cache.go
```diff
@@ -158,6 +159,11 @@ func (c *Cache) Create(ctx context.Context, game *entity.Game) error {
 	return c.setOrCreate(ctx, game, true)
 }
 
+// Create creates the game in the store only.
```
domino14, line 162: do you mean `CreateRaw`?

rpc/api/proto/mod_service/mod_service.twirp.go
```diff
@@ -1,6 +1,13 @@
-// Code generated by protoc-gen-twirp v8.1.1, DO NOT EDIT.
```
domino14, line 1: seems like we downgraded back to 8.0.0 -- can you re-generate your pb_compiler docker container?

pkg/stores/game/cache.go
jvc56: Oops, yes

go.sum
```diff
@@ -394,20 +1555,89 @@ gorm.io/datatypes v0.0.0-20200924071644-3967db6857cf/go.mod h1:n2DTgk9at7cr/CWOT
...
+modernc.org/zappy v1.0.0/go.mod h1:hHe+oGahLVII/aTTyWK/b53VDHMAGCBYYeZ9sn83HC4=
+rsc.io/binaryregexp v0.2.0/go.mod h1:qTv7/COck+e2FymRvadv62gMdZztPaShugOCi3I+8D8=
+rsc.io/pdf v0.1.1/go.mod h1:n8OzWcQ6Sp37PL01nO98y4iUCRdTGarVfzxY20ICaU4=
+rsc.io/quote/v3 v3.1.0/go.mod h1:yEA65RcK8LyAZtP9Kv3t0HmxON59tX3rD+tICJqUlj0=
+rsc.io/sampler v1.3.0/go.mod h1:T1hPZKmBbMNahiBKFy5HrXp6adAjACjK9JXDnKaTXpA=
+sigs.k8s.io/apiserver-network-proxy/konnectivity-client v0.0.14/go.mod h1:LEScyzhFmoF5pso/YSeBstl57mOzx9xlU9n85RGrDQg=
+sigs.k8s.io/apiserver-network-proxy/konnectivity-client v0.0.15/go.mod h1:LEScyzhFmoF5pso/YSeBstl57mOzx9xlU9n85RGrDQg=
+sigs.k8s.io/structured-merge-diff/v4 v4.0.2/go.mod h1:bJZC9H9iH24zzfZ/41RGcq60oK1F7G282QMXDPYydCw=
+sigs.k8s.io/structured-merge-diff/v4 v4.0.3/go.mod h1:bJZC9H9iH24zzfZ/41RGcq60oK1F7G282QMXDPYydCw=
+sigs.k8s.io/yaml v1.1.0/go.mod h1:UJmg0vDUVViEyp3mgSv9WPwZCDxu4rQW1olrI1uml+o=
```
domino14, line 1642: :saddest_pepe:

domino14:
- Need to pack migration files with the executable in Docker
- Need to break out PG host/password/username/etc into separate variables and create URL + standard conn strings

pkg/stores/puzzles/db.go
```diff
@@ -188,11 +187,79 @@ func (s *DBStore) GetPuzzle(ctx context.Context, userUUID string, puzzleUUID str
...
+			return "", err
+		}
+	}
+	if err != nil {
+		return "", err
+	}
+
+	if previousPuzzleUUID == "" {
+		var previousPid int
+		err = tx.QueryRowContext(ctx, `SELECT puzzle_id FROM puzzle_attempts WHERE updated_at <= $1 AND puzzle_id != $2 ORDER BY updated_at DESC LIMIT 1`, currentPuzzleUpdatedTime, pid).Scan(&previousPid)
```
domino14, line 242: should we add an index on `updated_at`?

pkg/stores/common/test_db.go
```diff
@@ -18,14 +18,11 @@ import (
...
 var MigrationsPath = os.Getenv("DB_MIGRATIONS_PATH")
-
-const (
-	TestDBName     = "liwords_test"
-	TestDBPort     = "5432"
-	TestDBUser     = "postgres"
-	TestDBPassword = "pass"
-	TestDBSSLMode  = "disable"
-)
+var TestDBName = os.Getenv("DB_NAME")
```
domino14, line 21: ! this will delete your local database when running tests, please change this env var. `liwords` is the DB_NAME, the test db name should be `liwords_test` most likely.

PR #945, opened 2022-03-17
bring in player-index refactor from Macondo + game ID + removal of secondWentFirst

domino14:
- use playerIndex instead of nickname in events
- liwords controls the short game ID and not macondo
- Remove secondWentFirst logic (backend should migrate in place when loading games from DB, for now).
- liwords should assign firsts explicitly (by passing the player array in order)
- ~~p1 and p2 in the game db row should correspond to the order of the players as well -- this needs to be migrated, along with the fields in QuickData, winnerIdx, etc.~~

Note on the last item above: we don't need to change the game DB row to have p1 and p2 correspond to the order. Those correspond to who made the request and who accepted (for non-tournament games).


domino14:
need to test this while swapping out backend.

domino14:
Local test before merging this in:

- Run bot version https://github.com/domino14/macondo/commit/9bc5ff48fd8a5c3297ec916017dcaca9bfd1d771 (see comment in https://github.com/domino14/macondo/pull/168) and liwords@master
- start a game
- check that second_went_first is True for maximum disruption, if not, keep making new games
- play a couple of turns, then kill the app
- checkout this branch and try to continue game.

domino14:
front-end breaks because there's no nicknames, until refresh. consider fixing this without requiring a refresh, if easy.

PR #981, opened 2022-04-20
Accept Vertical Opening Puzzle Answers

jvc56:


pkg/puzzles/puzzles.go
```diff
@@ -249,13 +248,37 @@ func answersAreEqual(userAnswer *ipc.ClientGameplayEvent, correctAnswer *macondo
...
+		userAnswer.Direction == correctAnswer.Direction) ||
+		(answerHitsCenterSquare(correctAnswer) &&
+			userAnswer.Row == correctAnswer.Column &&
+			userAnswer.Column == correctAnswer.Row &&
+			userAnswer.Direction != correctAnswer.Direction)
+}
+
+func answerHitsCenterSquare(answer *macondopb.GameEvent) bool {
+	var startSquare int32
+	if answer.Row == 7 && answer.Column <= 7 && answer.Direction == macondopb.GameEvent_HORIZONTAL {
```
domino14, line 268: if we have ZOMGWords puzzles we'll have to not hardcode these numbers. I don't see a straightforward way to pass board dimensions all the way down to here though. It's probably ok for now.

PR #985, opened 2022-04-23
958 keep lexicon filter

jonager:


domino14:
Thanks! If we have more than two repeats it may be worth factoring this code out into its own React hook or similar, but this is ok for now.

liwords-ui/src/lobby/active_games.tsx
```diff
@@ -17,12 +24,44 @@ type Props = {
...
+            }
+          }
+        } else {
+          // filter is reset, remove lexicon
+          setSortByLexicon(null);
+          localStorage.removeItem('sortByLexicon');
+        }
+      }
+    },
+    []
```
domino14, line 62: this requires a dependency `sortByLexicon` in this array, since if that changes, it should change the callback. The `docker-compose` logs for the `frontend` should show this.

liwords-ui/src/lobby/sought_games.tsx
```diff
@@ -124,6 +134,34 @@ export const SoughtGames = (props: Props) => {
...
+            }
+          }
+        } else {
+          // filter is reset, remove lexicon
+          setSortByLexicon(null);
+          localStorage.removeItem('sortByLexicon');
+        }
+      }
+    },
+    []
```
domino14, line 162: also needs the `sortByLexicon` in this array

liwords-ui/src/lobby/active_games.tsx
```diff
@@ -17,12 +24,44 @@ type Props = {
 };
 
 export const ActiveGames = (props: Props) => {
+  const { useState } = useMountedState();
+  const [sortByLexicon, setSortByLexicon] = useState(
+    localStorage.getItem('sortByLexicon')
```
domino14, line 29: could we rename the item inside localStorage to `lobbyFilterByLexicon` just to give it a little more of a namespace?

domino14:
Looks good overall; just a couple of hooks dependencies need to be resolved. If we want to use this code in more than these two places it may be worth it making a hook out of it, but that's ok for now.

liwords-ui/src/lobby/active_games.tsx
jonager: thanks, I'll add it, I haven't used react since 2017, so much has changed. I need to read about this

jonager:
> 

do we have any examples of something like this somewhere else in the project?

domino14:
> > 
> 
> do we have any examples of something like this somewhere else in the project?

We use React hooks throughout. You can see for example `liwords-ui/src/puzzles/puzzle.tsx` and search for `useEffect` there to see the dependencies of the various functions. But this is just one example of many throughout the React codebase.

PR #990, opened 2022-04-25
Refactor tournament ready

domino14:


domino14:
this probably will not be refactored in this way since we figured out #998. let's close it for now.

PR #1005, opened 2022-05-06
Next Puzzle by Closest Rating

jvc56:
We might be able to make the database query a bit better, but I'm not totally sure. Query inspired by [this stackoverflow answer](https://stackoverflow.com/questions/6102824/is-there-a-postgres-closest-operator).


domino14:
@jvc56 also need to resolve conflicts

PR #1019, opened 2022-05-21
Get next puzzle by rating

atomiccupcake:
postProto is sending a get instead of a post locally. holp.

image


atomiccupcake:
image

Adds a more readable, consistent view of ratings to the puzzle page, now gets puzzle by rating. As above, the endpoint fails completely if we run out of puzzles (will not load finished or skipped puzzles). I believe it should get old ones after that, like the other endpoint. Up to you whether you want to wait for a more graceful behavior or assume we can be sure to never get into this state in prod...

domino14 (approved):

domino14 (approved):

PR #1021, opened 2022-05-22
Profile shortcut

atomiccupcake:
Creates a shortcut to a user's own profile at /profile


liwords-ui/src/profile/profile.tsx
```diff
@@ -238,105 +244,111 @@ export const PlayerProfile = React.memo(() => {
...
-        setAvatarUrl(resp.data.avatar_url);
-        setAvatarsEditable(resp.data.avatars_editable);
-        setBio(resp.data.about);
-        setBioLoaded(true);
-      })
-      .catch((e: AxiosError) => {
-        setUserFetched(true);
-        errorCatcher(e);
-      });
+    if (username) {
```
domino14, line 251: let's do `if (!username) { return }` in this and the other one to make the diff a little nicer?

PR #1024, opened 2022-05-31
Puzzle widget

atomiccupcake:
PR 1019 has been merged into this (as it's dependent on that work), so the diff will get quite a bit simpler after that gets to master. You really won't want to review this until it is.

Adds a front page widget to show the next puzzle. 
image

This branch also _heavily_ reduces redundant css, by cleaning up our sass imports and migrating them to "use". 

Settings page has been updated to use the updated BoardPreview component with tile choices and board choices in the same graphic. This component (non-interactive static board with tiles) creates a nice path toward a wall of watchable live games...


liwords-ui/src/puzzles/puzzle_preview.tsx
```diff
@@ -0,0 +1,188 @@
...
+  const [userRating, setUserRating] = useState<number | undefined>(undefined);
+  const [puzzleRating, setPuzzleRating] = useState<number | undefined>(
+    undefined
+  );
+
+  const loadNewPuzzle = useCallback(async () => {
+    if (!userLexicon) {
+      // A hard coded id for a simple puzzle to display. Clicking in will
+
+      setPuzzleID('E3kGXKyzhYirNzsMfCW3QV');
```
domino14, line 48: we'll need to find a good puzzle from prod?

domino14: nm this is the prod puzzle lols

domino14 (approved):

PR #1025, opened 2022-06-06
reset filter when you creat a seek in a filtered out language

jonager:
@domino14 I'm able to reset `lobbyFilterByLexicon` in the state and localStorage when I create a new seek, but active games and sought games components are not updating, they still show the filter. If I change between the taps Play and Watch it updates and goes away. For some reason the change in the store is not triggering a rerender

### DIFF
--- a/liwords-ui/src/lobby/active_games.tsx
+++ b/liwords-ui/src/lobby/active_games.tsx
@@ -14,8 +14,11 @@ import { ActiveGame } from '../store/reducers/lobby_reducer';
 import { calculateTotalTime } from '../store/constants';
 import { VariantIcon } from '../shared/variant_icons';
 import { MatchLexiconDisplay } from '../shared/lexicon_display';
-import { useLoginStateStoreContext } from '../store/store';
-import { useMountedState } from '../utils/mounted';
+import {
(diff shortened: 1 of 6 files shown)


domino14:
Sorry for the long delay here.

What happens in this branch?

```javascript
    case ActionType.setLobbyFilterByLexicon: {
```

Do you know if it actually calls that? Have you put a log on it?

liwords-ui/src/lobby/active_games.tsx
```diff
@@ -40,9 +43,9 @@ export const ActiveGames = (props: Props) => {
 
   const handleChange = useCallback(
     (
-      pagination: TablePaginationConfig,
+      _pagination: TablePaginationConfig,
       filters: Record<string, FilterValue | null>,
-      sorter:
+      _sorter:
```
domino14, line 48: would like to understand a little bit better why these are underscored -- `_pagination` and `_sorter`.

domino14 (commented):
just have a question

liwords-ui/src/lobby/active_games.tsx
jonager: becuase it's not being used.

PR #1030, opened 2022-06-12
Remove more gorm

jvc56:


pkg/stats/stats.go
```diff
@@ -132,12 +134,13 @@ func InstantiateNewStats(playerOneId string, playerTwoId string) *entity.Stats {
...
 }
 
-func AddGame(stats *entity.Stats, lss ListStatStore, history *pb.GameHistory, req *ipc.GameRequest,
+func AddGame(ctx context.Context, stats *entity.Stats, lss ListStatStore, history *pb.GameHistory, req *ipc.GameRequest,
 	cfg *macondoconfig.Config, gameEndedEvent *ipc.GameEndedEvent, gameId string) error {
 	// Josh, plz fix these asinine calls to incrementStatItem
 	events := history.GetEvents()
 
-	info := &IncrementInfo{Cfg: cfg,
+	info := &IncrementInfo{Ctx: ctx,
```
domino14, line 142: In general you don't want to put context in structs. 

https://go.dev/blog/context-and-structs

However, this article highlights a case where you'd have to change a ton of functions and says it's probably ok to do that. I think as long as it's not abused it's probably ok.

pkg/stores/soughtgame/db.go
```diff
@@ -2,292 +2,353 @@ package soughtgame
...
-		return false, result.Error
+	tx, err := s.dbPool.BeginTx(ctx, common.DefaultTxOptions)
+	if err != nil {
+		return false, err
 	}
-	return count > 0, nil
+	defer tx.Rollback(ctx)
+
+	var exists bool
+	err = tx.QueryRow(ctx, `SELECT EXISTS(SELECT 1 FROM soughtgames WHERE seeker = $1)`, userID).Scan(&exists)
```
domino14, line 314: 🤯

domino14 (approved):

pkg/stores/soughtgame/db.go
jvc56: I don't see it. What did I mess up here?

domino14: No nothing, I just thought it was a cool query.

jvc56: :sunglasses:

PR #1038, opened 2022-07-02
Silent Mode backend profile setting/migration

domino14:


domino14:
This is ready for front-end @atomiccupcake 

Private chat has been disabled if we are a child or silent mode is on.

Chat in a "game" channel is disabled if we are a child/silent, but not if our opponent is a child/silent, as that involves some extra database lookups. However, the opponent's profile will show them as a child/silent, and we can just not show the chat at all if our opponent is silent (or show a message saying so)?

PR #1045, opened 2022-07-27
Attempt to get zomgwords working. almost there!

domino14:
These all must work before deploy

- [x] bot
- [x] stats
- [x] examine interface (size is now correct)
- [x] analyzer-wasm (@andy-k )
- [x] export gcg (should work but test it)
- [x] export image
- [x] "variants" dropdown in secret menu (rename from "wordsmog")
- [x] variant icon needs to show up in watch window, in profile cards/history.etc
- [x] colors for all themes
- [x] telestrator
- [x] blindfold mode (currently `;p1` is invalid, `;a22` throws console error, `;a21` works)


domino14:
@atomiccupcake I think I need help. The tile size, font size, etc are hard-coded in `base.scss` and the 21x21 board doesn't show the tiles in the right place or at the right size:

image

domino14:
to get stats working must fix `countBonusSquares` function

PR #1052, opened 2022-07-31
fix zomgwords board layout

andy-k:
note: https://github.com/domino14/macondo/pull/177 has a CI issue. if it is updated, this should be updated too


andy-k:
(analyzer is already correct)

PR #1065, opened 2022-09-11
add an IRL mode hack for tournaments

domino14:


pkg/tournament/classic_division.go
```diff
@@ -406,11 +407,41 @@ func (t *ClassicDivision) SubmitResult(round int,
 
 	pk1, err := t.getPairingKey(p1, round)
 	if err != nil {
-		return nil, err
+		found := false
```
jvc56, line 410: Shouldn't we check that we are actually in IRL mode at some point in this if block?

pkg/tournament/classic_division.go
```diff
@@ -406,11 +407,41 @@ func (t *ClassicDivision) SubmitResult(round int,
...
+		}
+		pk1, err = t.getPairingKey(p1, round)
+		if err != nil {
+			return nil, err
+		}
 	}
 	pk2, err := t.getPairingKey(p2, round)
 	if err != nil {
-		return nil, err
+		found := false
```
jvc56, line 429: Same as above.

pkg/tournament/classic_division.go
domino14: yeah, you're probably right.

PR #1074, opened 2022-10-31
set chat tab as default for game channel

jonager:


liwords-ui/src/chat/chat.tsx
```diff
@@ -329,10 +329,10 @@ export const Chat = React.memo((props: Props) => {
   }, [defaultChannel, defaultDescription, loggedIn, props.suppressDefault]);
 
   useEffect(() => {
-    if (props.tournamentID) {
+    if (props.tournamentID || channelType === 'game') {
```
domino14, line 332: let's add `gametv` as a possible channel type for this too

domino14 (changes requested):
let's do it for gametv too (for observers of ongoing games)

liwords-ui/src/chat/chat.tsx
jonager: done, maybe it's a good idea to create a channelType enum if we start using it in other places

domino14 (approved):

domino14:
merged copy branch #1076

PR #1084, opened 2022-11-12
Board editor

domino14:


domino14:
- analyzer sometimes breaks and only shows Pass as an option
- game didn't end after final play
- should add ability to edit previous racks
- should add place-move ability to analyzer

PR #1093, opened 2022-12-09
GameDocument and cwgame library

domino14:
- Add cwgame gameplay library. This will eventually replace the Macondo dependency. It is a simple library that relies upon an in-memory representation of a game, which we call a GameDocument
- A GameDocument  contains all the information necessary to play an OMGWords game, including the current state of the board, bag, timers, racks, events, etc. It should be quick to serialize / deserialize (need benchmarks) and should eventually obviate the need for a game cache.


domino14:
Before merging, **must set DATA_PATH ENVIRONMENT VARIABLE ON PROD**

PR #1107, opened 2023-01-21
Editor improvements

domino14:


domino14:
- [ ] should allow overwriting of saved game on database if that's what the user really wants

PR #1108, opened 2023-01-21
fix blocking logic

jonager:


pkg/bus/meseeks.go
```diff
@@ -148,9 +148,11 @@ func (b *Bus) newSeekRequest(ctx context.Context, auth, userID, connID string,
 		if err != nil {
 			return err
 		}
-		err = checkForBlock(ctx, b, connID, requester, receiver)
+		blockingState, err := checkForBlock(ctx, b, connID, requester, receiver)
 		if err != nil {
 			return err
+		} else if blockingState == 0 || blockingState == 1 {
+			return nil
```
jonager, line 155: @domino14 I'm not sure if I should return a new error here instead

domino14: that's a good question. It might be better to leave it as is, just with no information about blocking.

domino14 (approved):

domino14:
this looks good to me. I'm going to run a couple tests and then deploy it. Maybe later we can write more tests for this functionality, which is definitely pretty untested.

pkg/bus/meseeks.go
```diff
@@ -699,10 +718,29 @@ func (b *Bus) openSeeks(ctx context.Context, receiverID string, tourneyID string
 	if len(sgs) == 0 {
 		return nil, nil
 	}
+
+	receiver, err := b.userStore.GetByUUID(ctx, receiverID)
```
andy-k, line 722: does this work when not logged in?
see #1138

pkg/bus/meseeks.go
```diff
@@ -699,10 +718,29 @@ func (b *Bus) openSeeks(ctx context.Context, receiverID string, tourneyID string
...
+	if err != nil {
+		return nil, err
+	}
+
 	log.Debug().Str("receiver", receiverID).Interface("open-matches", sgs).Msg("open-matches")
 	pbobj := &pb.SeekRequests{Requests: []*pb.SeekRequest{}}
 	for _, sg := range sgs {
-		pbobj.Requests = append(pbobj.Requests, sg.SeekRequest)
+		seeker, err := b.userStore.GetByUUID(ctx, sg.SeekRequest.User.UserId)
```
andy-k, line 730: is there a way to get multiple (and possibly combine this with the receiver Get above)? otherwise each is possibly a separate roundtrip to db

andy-k: or maybe have ListOpenSeeks do an antijoin so that in the first place it doesn't return games that would be filtered off and we don't even have to load the full user info of everyone currently playing?

pkg/bus/meseeks.go
jonager: don't you have to be logged in to play? I thought we didn't allow anonymous play

andy-k: - for lobby, from `initRealmInfo`, it seems [`sendLobbyContext`](https://github.com/domino14/liwords/blob/891b10ba4e7a4337225ddec6121256734d11ba5b/pkg/bus/bus.go#L1024) sends `openSeeks` and then `activeGames` even if `userIsAnon`
- `openSeeks` fails (an ephemeral random userid is assigned for each anonymous tab, doing `GetByUuid` on these returns "user not found"), so `activeGames` isn't sent out, possibly causing #1138

(frontend doesn't currently show `openSeeks` result, but seems to maintain it internally anyway)

andy-k: (you do have to be logged in to play, but not just to see list of active games, watch them, and possibly see list of seeks)

[if there's no sought games](https://github.com/domino14/liwords/blob/891b10ba4e7a4337225ddec6121256734d11ba5b/pkg/bus/meseeks.go#L718) `GetByUUID` is skipped and the anonymous user can see list of active games. (hence it's hard to see this happening locally)

jonager: Thank you for the explanation, yes,  I didn't see this issue in my local. I'll try to fix it

PR #1122, opened 2023-02-05
do not show dropdown for not anon users in the player"s profile

jonager:


domino14 (approved):

liwords-ui/src/shared/usernameWithContext.tsx
```diff
@@ -180,7 +180,7 @@ export const UsernameWithContext = (props: UsernameWithContextProps) => {
       overlay={userMenu}
       getPopupContainer={() => document.getElementById('root') as HTMLElement}
       placement="bottomLeft"
-      trigger={['click']}
+      trigger={!loggedIn && props.omitProfileLink ? [] : ['click']}
```
andy-k, line 183: breaks https://woogles.io/profile/HastyBot (#891) when not logged in.

also, mouseover cursor is still suggesting it's clickable.

PR #1222, opened 2023-11-06
Header for tied game hard to read in dark mode

emma-britton:
I finally had a game end in a tie. The text is difficult to read because the background is so light.
I changed it to a darker colour available within the colour scheme.

(This is a more "proper" fix that defines separate colours for the profile game headers instead of re-using the timer colours. This means it can be kept the same in light mode, where it look OK, and only changed in dark mode.)

Capture

### DIFF
--- a/liwords-ui/src/profile/profile.scss
+++ b/liwords-ui/src/profile/profile.scss
@@ -216,20 +216,20 @@
     }
     .ant-card-head {
       @include colorModed() {
-        background-color: m($timer-low-light);
+        background-color: m($profile-game-tie);
       }
     }
     &.win {
       .ant-card-head {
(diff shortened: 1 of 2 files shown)


liwords-ui/src/profile/profile.scss
```diff
@@ -216,20 +216,20 @@
...
-          background-color: m($timer-light);
+          background-color: m($profile-game-win);
         }
       }
     }
     &.loss {
       .ant-card-head {
         @include colorModed() {
-          background-color: m($timer-out-light);
+          background-color: m($profile-game-loss;
```
domino14, line 232: @emma-britton this needs a closing parenthesis

PR #1338, opened 2024-05-26
consolidate stores, fix #632

### LINKED ISSUE #632: Consolidate Stores
We should weto this:

`return gameplay.TimedOut(ctx, b.gameStore, b.userStore, b.notorietyStore, b.listStatStore, b.tournamentStore, evt.UserId, evt.GameId)`

And replace it with something like this:

`return gameplay.TimedOut(ctx, gameplayStores, evt.UserId, evt.GameId)`

In all places where the stores are getting out of hand.

### LINKED ISSUE #1314: delete user cache code

domino14:
fix #632 
fix #1314


domino14:
This doesn't quite fix the fact that we can't really do transactions etc. Obviously this needs to be reworked further. But it should fix a lot of ugliness around function signatures.

PR #1377, opened 2024-10-06
Pairings API

jvc56:
Creates the Pairings API which will initially only support COP pairings. COP pairings are currently implemented for TSH as a [Perl module](https://github.com/jvc56/tournament_pairing_algorithms/blob/main/COP.pm). I would like to migrate COP to Woogles for several reasons:

- Make COP pairings accessible beyond TSH.
- Make COP pairings faster by implementing them in Go instead of Perl.
- Rewrite the COP implementation to improve readability and maintainability now that the COP algorithm has matured.

COP pairings require a nontrivial amount of compute. Which means if we made this as an API we would have to think about:

- How to execute pairing requests that require compute (as a lambda job maybe?) and the max runtime for such requests.
- How to authenticate requests that are made programatically (with some sort of token maybe?) and if requests should be rate limited somehow.
- If authentication is done with tokening, how to distribute, store, expire. etc. tokens.


jvc56:
The implementation of COP is complete. It is much faster than I had originally anticipated. We can likely keep all request to <10 seconds. There were many inefficiencies in the original Perl implementation that I didn't realize until now. The bottleneck for the sims is sorting the standings after each simulated round. There might also be more efficient ways to compute control loss that require fewer sims.

Issue #1395, opened 2024-12-02
Patreon integration w/BestBot


domino14:
- [x] Full integration with Patreon
- [x] Implement new design for selecting BestBot
- [x] Implement integrations page
- [x] Enable paywall for BestBot
- [x] Possibly insert an override/flag for some users

domino14:
#1423

PR #1397, opened 2024-12-10
Bestbot patreon paywall

domino14:
#1395


domino14:
The paywall code is there but not enabled until we go live.

PR #1402, opened 2024-12-13
show solved puzzle solution in dev mode

### LINKED ISSUE #1400: solved puzzles don't show their placed tiles in dev mode
only when turning StrictMode off. the entire app is useEffect spaghetti

andy-k:
fixes #1400


andy-k:
StrictMode's second run of `BoardPanel`'s `fullReset` did it, so `placeGameEvt(puzzleInfo.solution);` needs to run after that instead of just before.

PR #1405, opened 2024-12-15
add blank prettierrc file to use default prettier settings

domino14:


andy-k (approved):

liwords-ui/.prettierrc
```diff
@@ -0,0 +1 @@
+{}
```
andy-k, line 1: \n

PR #1435, opened 2025-02-12
Set priority for routes in docker-compose

BenMusch:
This seems to be the most idiomatic workaround for the issue I found where all requests routed to the front-end service


docker-compose.yml
BenMusch, line 1: Seems like my auto-formatter ran on this file. Let me know if you want me to revert. The only changes are the "priority" settings

BenMusch:
**[dont merge]**

hitting some issues locally just need to confirm this is working as expected. Will comment again once I do

PR #1436, opened 2025-02-12
Make nudge-restrictions aware of overtime

BenMusch:
Closes https://github.com/woogles-io/liwords/issues/1394

Allows nudges if (and only if) the total time remaining for the nudged player is > 2 minutes, including overtime. This aligns with the amount of time a user has to dismiss the nudge

Note that this does not account for increment, we can maybe introduce a heuristic for that based on increment size and tiles remaining if desirable.

### DIFF
--- a/pkg/gameplay/game_playing_test.go
+++ b/pkg/gameplay/game_playing_test.go
@@ -69,7 +69,27 @@ func (ec *evtConsumer) consumeEventChan(ctx context.Context,
 	}
 }
 
-func makeGame(cfg *config.Config, stores *stores.Stores) (
+type TestGameOption func(*pb.GameRequest)
+
+func WithIncrementSeconds(seconds int32) TestGameOption {
+	return func(gr *pb.GameRequest) {
+		gr.IncrementSeconds = seconds
+	}
+}
+
+func WithMaxOvertimeMinutes(minutes int32) TestGameOption {
+	return func(gr *pb.GameRequest) {
+		gr.MaxOvertimeMinutes = minutes
+	}
+}
+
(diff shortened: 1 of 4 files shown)


pkg/gameplay/game_playing_test.go
```diff
@@ -69,7 +69,27 @@ func (ec *evtConsumer) consumeEventChan(ctx context.Context,
 	}
 }
 
-func makeGame(cfg *config.Config, stores *stores.Stores) (
+type TestGameOption func(*pb.GameRequest)
+
+func WithIncrementSeconds(seconds int32) TestGameOption {
```
BenMusch, line 74: Claude suggested this pattern, it's other suggestion was:


```go
type GameConfig struct {
    IncrementSeconds   int32
    MaxOvertimeMinutes int32
}

func makeGame(cfg *config.Config, stores *stores.Stores, gameCfg *GameConfig) (
    *entity.Game, *entity.FakeNower, context.CancelFunc, chan bool, *evtConsumer) {
    
    // ... same setup ...
    gr := proto.Clone(gameReq).(*pb.GameRequest)

    if gameCfg == nil {
        gr.IncrementSeconds = 5
        gr.MaxOvertimeMinutes = 0
    } else {
        gr.IncrementSeconds = gameCfg.IncrementSeconds
        gr.MaxOvertimeMinutes = gameCfg.MaxOvertimeMinutes
    }
    
    // ... rest of function ...
}
```

lmk if you'd prefer that approach (or some other option)

domino14: I like the pattern you're using, it is common in Go.

domino14 (approved):

PR #1439, opened 2025-02-16
[tournaments] Support per-round spread cap

BenMusch:
Back-end implementation of a per-round spread cap. The idea here is to provide an _override_ at the round level, so it's clear that we fall back to the tournament-level setting when no override is specified


gen.sh
```diff
@@ -5,7 +5,12 @@ set -e
 RPC_PATH="rpc"
 MODELS_PATH="pkg/stores/models"
 UI_GEN_PATH="liwords-ui/src/gen"
-USER_GROUP="unprivileged:unprivileged"
+
+if [[ "$OSTYPE" == "darwin"* ]]; then
```
BenMusch, line 9: This was necessary to get this script to work on mac OS

api/proto/ipc/tournament.proto
```diff
@@ -112,6 +112,11 @@ message RoundControl {
   int32 repeat_relative_weight = 9;
   int32 win_difference_relative_weight = 10;
   bool interleave_team_round_robin = 11;
+  int32 spread_cap_override = 12;
```
BenMusch, line 115: self-review: I need to make this a uint

api/proto/ipc/tournament.proto
```diff
@@ -112,6 +112,11 @@ message RoundControl {
...
+  // We use this boolean to differentiate between a disabled override
+  // (i.e. using the division default) and an enabled override set to 0
+  // (i.e. overriding the division default to remove the cap)
+  bool spread_cap_override_enabled = 13;
```
BenMusch, line 119: Without optionals, `0` could potentially represent both:
 - Override the tournament-level cap and set the cap to `0` (i.e. remove the cap if it exists)
 - No round-override

Other options here are:
 - making `spread_cap_override` optional. Opted against this since code search shows no other optional fields in the entire project
 - allowing negatives, where `-1` means no override
 - Disallow either: setting a spread cap override when no tournament-level cap is set _or_ having an override which removes the spread-cap, therefore making `0` only represent one possible situation

I think we should either allow optionals in this instance or keep this boolean + int set-up, the other two are unappealing to me. But wanted to flag all the options

api/proto/ipc/tournament.proto
BenMusch, line 1: Just to check my understanding, do I need to add anything for the work here to get persisted anywhere, or is there some part of the code which will just persist all of the in-memory state whenever relevant? I see that the `tournament` table has a json field which does store this, but wanna make sure I'm not missing a place to specifically persist this new field

api/proto/ipc/tournament.proto
```diff
@@ -121,7 +126,7 @@ message DivisionControls {
   TournamentGameResult suspended_result = 4;
   int32 suspended_spread = 5;
   bool auto_start = 6;
-  int32 spread_cap = 7;
+  uint32 spread_cap = 7;
```
BenMusch, line 129: drive-by change, happy to revert. I dont think negatives are valid here

api/proto/ipc/tournament.proto
domino14: I don't have to be super dogmatic about optionals; it does make sense in this case to have optionals. I'll say I would also be ok with a spread cap of 0 meaning no spread cap, since it seems we might never want a spread cap of 0 (what would it mean? every game is a binary win/loss? I guess it makes some sense).

I do prefer an optional over having two variables for it. 

Looking at this now:

https://protobuf.dev/programming-guides/field_presence/

It seems that the protobuf people changed their minds again about explicit vs implicit presence. So now they recommend `optional` again (it wasn't in proto3 for the longest time, then experimental, and now they recommend it?).

PR #1440, opened 2025-02-17
[tournaments] Add start/end date to tournaments + GetRecentAndUpcomingTournaments

BenMusch:
This is the first step towards listing tournaments on the home page. We need to be able to look up tournaments by when they start/end, and this feels more natural than dumping it into the JSON.

My idea for how to structure this change:
1. Add start/end dates in the DB + API to query them
2. Start allowing tournaments to specify start/end dates via UI
3. Start showing the tournaments, queried by date/time, in the UI

Because the scale of tournaments is quite small, I figure we can just query based on timing and then filter for tournament type/status on the front-end.


pkg/stores/tournament/cache.go
```diff
@@ -27,6 +27,7 @@ type backingStore interface {
 	RemoveRegistrants(ctx context.Context, tid string, userIDs []string, division string) error
 	RemoveRegistrantsForTournament(ctx context.Context, tid string) error
 	ActiveTournamentsFor(ctx context.Context, userID string) ([][2]string, error)
+	GetRecentAndUpcomingTournaments(ctx context.Context) ([]*entity.Tournament, error)
```
BenMusch, line 30: This is a super simple API that's basically hard-coded for the "show some tournaments on the homepage" use-case.

I'm pretty open to other suggestions here which may be more extensible in the future

db/migrations/202502161828_tournament_scheduled_start_and_end.up.sql
```diff
@@ -0,0 +1,10 @@
+BEGIN;
+
+ALTER TABLE  tournaments
+ADD COLUMN scheduled_start_time TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT '1970-01-01 00:00:00+00',
+ADD COLUMN scheduled_end_time TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT '1970-01-01 00:00:00+00';
```
BenMusch, line 5: Defaults here basically default exclude anything created before this, which is fine. In the UI we can show a message like `No date available` is date.year < `2000`

domino14:
lgtm!

domino14 (approved):

PR #1441, opened 2025-02-18
[tournaments] Add start/end time to admin and tournament director forms

BenMusch:
Video walkthrough **(note: i say some incorrect things about this removing the need for the "merge" logic in the set metadata api. ignore that)**

<div>
    <a href="https://www.loom.com/share/bb1d8fad40954908bcc6ef5faba15d01">
      <p>Woogles.io - 17 February 2025 - Watch Video</p>
    </a>
    <a href="https://www.loom.com/share/bb1d8fad40954908bcc6ef5faba15d01">
      
    </a>
</div>

This PR adds the ability to set start+end time from the tournament management pages for admins and tournament directors. There's some unrelated code changes that I'll flag inline


liwords-ui/src/admin/tourney_editor.tsx
```diff
@@ -121,17 +130,20 @@ const SettingsModalForm = (mprops: {
 };
 
 export const TourneyEditor = (props: Props) => {
-  const [description, setDescription] = useState("");
```
BenMusch, line 124: This code was kinda an anti-pattern IMO. For react forms, you have a choice between a "controlled" form where you manage all of the `onChange` events, set the state locally to your component, and then bind the `value` of that input to the state. _or_ you can have an uncontrolled component where the state is managed externally

Here we were doing both, so there were essentially two sources of truth for field values: one in the local state of this component and one within the `Form.useForm()` hook state. I've removed the controlled state entirely, it shouldn't affect any behavior and just kills some dead code

liwords-ui/src/admin/tourney_editor.tsx
```diff
@@ -121,17 +130,20 @@ const SettingsModalForm = (mprops: {
...
-  const [name, setName] = useState("");
-  const [color, setColor] = useState("");
-  const [logo, setLogo] = useState("");
   const [settingsModalVisible, setSettingsModalVisible] = useState(false);
   const [selectedGameRequest, setSelectedGameRequest] = useState<
     GameRequest | undefined
   >(undefined);
   const [form] = Form.useForm();
+
+  const description = useWatch("description", form);
```
BenMusch, line 139: Since we're not using controlled state anymore, this hook lets us watch for changes to the fields we care about for the sidebar preview. Otherwise it doesn't update as users type

liwords-ui/src/admin/tourney_editor.tsx
```diff
@@ -221,21 +253,6 @@ export const TourneyEditor = (props: Props) => {
       flashError(err);
     }
   };
-  const onDescriptionChange = (evt: React.ChangeEvent<HTMLTextAreaElement>) => {
```
BenMusch, line 224: this and the on-change removals are cleaning up the controlled state

liwords-ui/.eslintrc
```diff
@@ -46,6 +46,13 @@
         "ignoreComments": true
       }
     ],
+    "no-restricted-imports": ["error", {
```
BenMusch, line 49: We do some necessary initialization of dayjs in the utils module, so we wanna make sure that nobody imports it without that initialization

liwords-ui/package.json
```diff
@@ -18,6 +18,7 @@
     "@types/react-dom": "^18.3.5",
     "antd": "^5.22.4",
     "browserslist-to-esbuild": "^2.1.1",
+    "dayjs": "^1.11.13",
```
BenMusch, line 21: this was already depended on by antd so it shouldn't increase the bundle or anything, just making the dependency explicit now that we actually use the library ourselves

domino14:
This is awesome, thanks!!

liwords-ui/src/admin/tourney_editor.tsx
domino14: thanks for fixing that; i am aware of the controlled vs uncontrolled but I must have totally brainfarted.

PR #1447, opened 2025-02-22
convert user uuids (first step)

domino14:
We should start using proper UUID fields instead of fields calling themselves UUIDs but that are actually the shortuuid TEXT encoding of a UUID. The latter ends up being bigger per column (22 vs 16 bytes) and indexing on it for queries like GetBriefProfiles (which happens very frequently) should be faster.


domino14:
this is probably dead in the water because of this: https://github.com/lithammer/shortuuid/issues/67

there's no way to convert some of our "uuids" back into the canonical 16-byte format.

PR #1449, opened 2025-02-25
[tournaments] Allow nullable tournament start+end date

BenMusch:
- Relax non-null constraint and remove default for these fields
 - Allow null values to be set (and cleared/added) from the UI


pkg/stores/tournament/db.go
```diff
@@ -159,7 +159,7 @@ func (s *DBStore) Set(ctx context.Context, tm *entity.Tournament) error {
 
 	ctxDB := s.db.WithContext(ctx)
 	result := ctxDB.Model(&tournament{}).Clauses(clause.Locking{Strength: "UPDATE"}).
-		Where("uuid = ?", tm.UUID).Updates(dbt)
+		Select("*").Where("uuid = ?", tm.UUID).Updates(dbt)
```
BenMusch, line 162: This is required to allow for the clearing of a field, otherwise the null value just gets dropped. Since the merge logic is handled within the `SetTournamentMetadata` endpoint, it seems to still behave as expected

domino14 (approved):

PR #1464, opened 2025-03-20
Fix firsts and seconds for team round robin

jvc56:


pkg/tournament/classic_division.go
```diff
@@ -1581,14 +1581,21 @@ func newClassicPairing(t *ClassicDivision,
 	firstMethod := t.RoundControls[round].FirstMethod
 
 	// Do not do this for Team RR, only RR
```
domino14, line 1583: is this comment obsolete?

jvc56: It sure is!

PR #1469, opened 2025-03-22
Attempt to fix firsts and seconds for shirts vs skins

jvc56:


domino14:
this was fixed with #1470

PR #1486, opened 2025-05-05
test a potential migration to GameDocument

domino14:


pkg/entity/utilities/document.go
```diff
@@ -162,25 +176,36 @@ func ToGameHistory(doc *ipc.GameDocument, cfg *config.Config) (*macondo.GameHist
 				rack = doc.Events[index+1].Rack
 			}
 		}
+		var tilesFromRack int
+		if evt.Type == ipc.GameEvent_EXCHANGE {
+			tilesFromRack = len(evt.Exchanged)
```
andy-k, line 181: does this handle unicode tiles?

domino14: it does actually. This event is a `GameDocument` event (`GameEvent`) not a Macondo `GameEvent`. The latter only deals with strings, the former deals with bytes and "`MachineLetters`".

domino14:
100000 games were tested on prod with no discrepancies. We can migrate games to S3 after translating to GameDocument. We can translate back to legacy GameHistory for now when loading old games.

PR #1505, opened 2025-08-17
trying to add an embeddable widget

domino14:


domino14:
this embed widget is nonworking, but we have some scaffolding setup for the backend of it, and this also updates FE libraries.

PR #1573, opened 2025-11-10
Filter My Correspondence Games section by user ID on front-end

BenMusch:
Fixes https://github.com/woogles-io/liwords/issues/1570

Filtered on the front-end since I believe we still want the game data broadcast to the general lobby of games, but happy to implement differently

### DIFF
--- a/liwords-ui/src/lobby/correspondence_games.tsx
+++ b/liwords-ui/src/lobby/correspondence_games.tsx
@@ -96,7 +96,6 @@ export const CorrespondenceGames = (props: Props) => {
 
   const formatGameData = useCallback(
     (games: ActiveGame[]): CorrespondenceGameTableData[] => {
-      // Filter to only show games where the current user is a participant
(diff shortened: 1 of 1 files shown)


domino14 (commented):
This definitely seems like it'll work but I am curious if you could track down what is causing these games to be sent to the frontend. Are they correspondence games as well? I feel like those shouldn't be sent at all if you're not the logged in user.

BenMusch:
@domino14 Ok yeah, confirmed with some logs:

 - On accept, game gets broadcast in `broadcastGameCreation()` of `meseeks.go`
 - On the front-end, the `ONGOING_GAME_EVENT` is received in `liwords-ui/src/store/socket_handlers.ts`. It then gets detected as correspondence, and there's no filtering by user ID. So maybe it makes more sense to filter on the event-handling in the front-end (e.g. the reducer/socket code rather than the component definition)

domino14:
Thanks! This makes sense to me now. I think your fix is fine for now.

PR #1606, opened 2025-11-26
refactor league scheduling: hourly runner, editable dates, new promo …

domino14:
…formula

Major changes:
- Add UpdateSeasonDates RPC to allow editing season start/end dates from admin
- Refactor league tasks from midnight/8am to hourly runner with idempotency
- Add task tracking columns (closed_at, started_at, etc.) to prevent duplicate runs
- Spread game creation over time with batching (150ms delay per 10 games)
- Change promotion/relegation formula from ceil(N/6) to ceil((N+1)/5)

The hourly runner uses actual season start/end dates instead of fixed times, allowing different leagues to start at different times (e.g., CSW at 8am UTC, NWL at 8am Eastern).

### DIFF
--- a/liwords-ui/src/leagues/admin.tsx
+++ b/liwords-ui/src/leagues/admin.tsx
@@ -32,7 +32,10 @@ import { flashError } from "../utils/hooks/connect";
 import { useLoginStateStoreContext } from "../store/store";
 import { SeasonStatus } from "../gen/api/proto/ipc/league_pb";
 import { ChallengeRule } from "../gen/api/proto/ipc/omgwords_pb";
-import { dayjsToProtobufTimestampIgnoringNanos } from "../utils/datetime";
+import {
+  dayjsToProtobufTimestampIgnoringNanos,
+  dayjs,
+} from "../utils/datetime";
 import { timestampDate } from "@bufbuild/protobuf/wkt";
 import { AllLexica } from "../shared/lexica";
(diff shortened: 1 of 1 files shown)


cmd/maintenance/main.go
```diff
@@ -88,6 +88,12 @@ func main() {
 		case "league-season-starting-soon":
 			err := LeagueSeasonStartingSoon()
 			log.Err(err).Msg("ran leagueSeasonStartingSoon")
+		case "league-hourly-runner":
+			err := LeagueHourlyRunner()
+			log.Err(err).Msg("ran leagueHourlyRunner")
+		case "league-unstarted-game-reminder":
```
domino14: can probably fold into the league hourly runner

db/queries/leagues.sql
```diff
@@ -65,6 +65,39 @@ UPDATE league_seasons
...
+SET divisions_prepared_at = NOW(), updated_at = NOW()
+WHERE uuid = $1;
+
+-- name: MarkSeasonStarted :exec
+UPDATE league_seasons
+SET started_at = NOW(), updated_at = NOW()
+WHERE uuid = $1;
+
+-- name: MarkRegistrationOpenedForNextSeason :exec
+-- Called on the CURRENT season to mark that registration for the NEXT season was opened
```
domino14: don't like this

PR #1634, opened 2025-12-10
add a maintenance overlay and pause real-time games when deploying

domino14:
## Need to expose `/ping` in Cloudfront to the front end prior to deploying this!


domino14:
Need to touch up some aspects of the UX around this. This is a fairly significant change.

PR #1700, opened 2026-02-26
return game end reason for league games

andy-k:


domino14:
plz merge main or rebase

PR #1870, opened 2026-06-20
add the australian draw tournament pairing method

andy-k:
## Summary
Adds the Australian Draw pairing method: a transparent, hand-playable, deterministic alternative to the matching-based methods. Round 1 is a casement draw (top half vs bottom half); every later round pairs the top remaining player with the highest-standing legal opponent that still leaves the rest of the field pairable, backtracking when a choice paints the draw into a corner, so a repeat-free draw is always found when one exists. A new 1-based `reset_round` round control (proto field 20) lets a director forgive meetings before a day boundary: a meeting in round r is avoided iff r >= reset_round (1 = avoid all prior, the default; >= current round = King-of-the-Hill). When a strict pass cannot pair the field the reset point is relaxed one round at a time, then errors cleanly.

- proto: `PairingMethod.AUSTRALIAN_DRAW = 12`, `RoundControl.reset_round = 20`; regenerated `tournament.pb.go` and `tournament_pb.ts`.
- `pkg/pair/australian.go`: casement (two-phase backtracker, zero top-vs-top whenever achievable), the later-round backtracking matcher + reset-relaxation retry, dispatcher; wired into `pkg/pair/pair.go`.
- `pkg/entity/pair.go`: `UnpairedPoolMembers.RepeatRounds` (per-round opponent history).
- `pkg/tournament/classic_division.go`: `getRepeatRounds` + `RepeatRounds` population for this method (round > 0).

## Odd fields
An odd field is padded with a phantom bye at the bottom of the standings and run through the even algorithm; whoever the phantom pairs takes the real bye. This unifies the odd and even cases with no special-casing -- the casement and the later-round matcher are both even-only.

## Guarantees
- Deterministic: standings sort, then nearest-first search; no randomness, hand-reproducible.
- Round 1 emits zero top-vs-top whenever a block-free draw with none exists, falling back to the minimal forced top-vs-top only when unavoidable.
- Round 2+ finds a repeat-free draw whenever one exists.
- All three guarantees are checked exhaustively against independent brute-force oracles (see test plan).

## Test plan
- [x] `go test ./pkg/pair/ -run TestAustralian` (34 tests, incl. exhaustive oracle cross-checks: casement even n=4,6 all block masks + n=8 sampled; casement odd n=3,5,7 all + n=9 sampled; later-round matcher n=2,4,6 all + n=8 sampled -- each vs an independent brute-force matcher)
- [x] `go test ./pkg/pair/` (full package)
- [x] `go test ./pkg/tournament/ -run TestClassicDivision`
- [x] `gofmt -l` clean on the changed Go files; `go build ./pkg/... ./cmd/...`

Merge/review this PR before #1871 -- the director-UI PR (#1871) depends on the generated `tournament_pb.ts` bindings added here and will not build on master until this lands.


api/proto/ipc/tournament.proto
```diff
@@ -133,6 +134,19 @@ message RoundControl {
...
+  // avoided as a repeat iff r >= reset_round, so meetings in earlier rounds
+  // are allowed to recur. This emulates a day-based reset (e.g. set 9 so
+  // rounds 1-8 are forgiven when pairing round 9 onward). 0/unset is treated
+  // as 1 (reset from round 1 = avoid all prior meetings, the default
+  // no-repeat behavior); do not rely on the proto3 zero-default as a
+  // meaningful sentinel -- the reader clamps it explicitly. reset_round at or
+  // past the current round forgives every prior meeting (King-of-the-Hill).
+  // When even a strict pairing is impossible the reset point is relaxed
+  // upward one round at a time.
+  uint32 reset_round = 20;
```
domino14, line 149: don't we need multiple reset rounds? i thought it resets every day.

PR #1983, opened 2026-09-19
End game on replayed final challenge; share going-out game-over helper

### LINKED ISSUE #1350: Game-ending challenge in uploaded game leaves board editor stuck in edit mode without saving
Uploaded GCG below, but unable to save, and every time I try to reenter board editor from that moment onward it takes me back to this game and I can no longer see the list of all uploaded games.

[Word Cup '24 Rd 28 Eta Karo.txt](https://github.com/user-attachments/files/16482678/Word.Cup.24.Rd.28.Eta.Karo.txt)
Screenshot 2024-08-03 at 14 52 20

magratheazaphod:
Fixes #1350.

A play that goes out and is then unsuccessfully challenged ends the game in `challengeEvent`, but `ReplayEvents` had no equivalent: it applied the challenge bonus and left the game in `WAITING_FOR_FINAL_PASS`. `ImportGCG` deliberately inserts no dummy pass here, so any GCG whose final play survived a challenge imported as a permanently unfinished game, which locks the board editor and blocks further imports.

Replay now ends the game at that point, using the same helper as the live path. The going-out player is taken from the placement that put the game in `WAITING_FOR_FINAL_PASS`, not from the challenge bonus event, since imported events are untrusted. A replayed prefix (editor amendments, `SetRacks`) is left unfinished, and the challenger stays on turn as in a live game.

Also folds the four going-out endings into one `endGameAfterGoingOut`, which fixes a live bug: `challengeEvent` never set `Winner`, so a player who went out was not recorded as the winner even when they finished ahead.


magratheazaphod:
Validation: ran the ImportGCG parse and replay steps over 30 real GCGs with this event shape (the pattern that hits this bug). Before: 0 of 30 reached GAME_OVER. After: 29 of 30, with final scores matching the GCG, including the file attached to the issue; the remaining one fails earlier on an invalid rack in its source.

Notes on the challenge rules: DOUBLE never reaches the new code, since it records a failed challenge as UNSUCCESSFUL_CHALLENGE_TURN_LOSS and ends via playMove. TEN_POINT shares the FIVE_POINT path, and TRIPLE returns from its own branch with Winner already set. Tests cover FIVE_POINT and SINGLE.

magratheazaphod:

Head SHA reviewed: `807e3c5c28ef4cf7f8df19c109851025bf7691e4`

**Verdict: READY TO PUBLISH** — no blockers.

### File classification

| File | Call | Reason |
| --- | --- | --- |
| `pkg/cwgame/api.go` | PUBLIC | core game engine |
| `pkg/cwgame/game.go` | PUBLIC | core game engine |
| `pkg/cwgame/api_test.go` | PUBLIC | unit tests, fixtures already in repo |

No secrets, infrastructure identifiers, site data, player data or machine-specific paths in the diff or in the three commit messages. No untracked files pending. No generated code (`rpc/`, `liwords-ui/src/gen/`, `pkg/stores/models/`), no migrations, no frontend and no SQL touched, so those checks do not apply.

### Findings

**SHOULD-FIX — `pkg/cwgame/api.go:436` — a replayed event still overwrites a player's score outright (pre-existing, not introduced here).**
The `default:` branch does `gdoc.CurrentScores[evt.PlayerIndex] = evt.Cumulative`, taking a cumulative score straight from a replayed event. For `ImportGCG` that value comes from an uploaded GCG, so a file that misreports a cumulative score silently sets a wrong score rather than failing. This PR deliberately stops *compounding* it: the end-rack points and the winner are now derived from replay state rather than from the event. Worth a follow-up to validate the cumulative against the engine's own running score and reject a mismatch; out of scope here.

**NIT — `pkg/cwgame/game.go:86` — the going-out scan stayed inline.**
`playMove` still finds the player who went out by scanning `gdoc.Racks` for the empty rack. Replay cannot use that (racks mid-replay are not reliable, which a run over real GCGs confirmed: the empty rack there can be the opponent's), so it tracks the going-out player from the placement instead. Two different derivations for the same question is a little awkward, but folding them together would be wrong — they genuinely have different information available.

### Tests

Both new tests fail on `master` and pass here, so the bug is genuinely covered:
- `TestReplayEventsChallengedGoingOutPlay` — plays a going-out move, challenges it, replays the resulting events, and checks the replay reaches the same state as the live game, including the on-turn player. Also covers a replayed prefix (must stay unfinished) and a tampered challenge-bonus player index (must not misassign end-rack points).
- `TestChallengeGoodWordEndOfGameSetsWinner` — covers the `Winner` fix in the live path.

Challenge-rule coverage: `FIVE_POINT` and `SINGLE` are tested. `DOUBLE` never reaches this code (a failed challenge there is an `UNSUCCESSFUL_CHALLENGE_TURN_LOSS`, ended by `playMove`), `TEN_POINT` shares the `FIVE_POINT` path, and `TRIPLE` returns from its own branch with `Winner` already set.

### Checks run

| Check | Result |
| --- | --- |
| `go build ./pkg/... ./cmd/...` | pass |
| `go build ./...` | fails only at the repo root, pre-existing and unrelated (`gen.go` is `package main` with no `main`) |
| `go vet ./pkg/cwgame/...` | pass |
| `go test ./pkg/cwgame/...` | pass |
| `gofmt -l pkg/cwgame` | clean |

Frontend checks (`tsc`, lint) not run: no frontend files changed. `pkg/omgwords` tests not run: they need a database.

PR #1986, opened 2026-09-21
Clear annotated done flag on reopen; disable delete for finished games

magratheazaphod:
Annotators hit "you cannot delete a game that is already done" with no warning. The policy stays as is; this makes it less surprising.

**UI:** the editor now shows a disabled "Delete this game" button with a tooltip ("Finished games can't be deleted.") when the game's play state is GAME_OVER, instead of offering a delete that the server will refuse.

**Sticky done flag:** `annotated_game_metadata.done` was set when a game ended (event or import) but never cleared. If an edit reopened the game, delete was still refused and the game was missing from `GetMyUnfinishedGames`. A new `ClearAnnotatedGameDone` store method is called when:
- `SendGameEvent` handles an event that leaves the game not over (amendments, or a challenge that takes back the final play),
- `SetRacks` leaves the game not over,
- `ReplaceGameDocument` stores a document that isn't GAME_OVER.

`PatchGameDocument` can't change game structure, so it's untouched. The clear is `WHERE done`, so it's a no-op write for games that were never done.

**Hooks / analysis:** `onGameDone` (analysis requests, broadcast `annotationDone`) is unchanged. It still fires whenever an event ends the game, which already happened on re-ending amendments before this change. Clearing the flag doesn't undo a previous analysis request or broadcast completion, and `ReplaceGameDocument` doesn't mark games done or fire the hook.

**Tests:** added `TestReopenedGameIsNotDone` (import finished GCG, replace with a non-over document, flag is cleared). It needs the test Postgres, so I couldn't run it locally; it compiles. Frontend: tsc and eslint pass.


magratheazaphod:
Closing for now; not ready to propose this yet. Branch kept.

PR #1987, opened 2026-09-21
Click a turn to jump to it in annotated and finished games

magratheazaphod:
Clicking a turn in the scorecard only seeked when already examining. On \`/anno/:gameID\` (and finished games) you are not examining on load, so clicks did nothing.

- \`ScoreCard\` takes an optional \`onSeek\` used when not examining.
- \`Table\` passes one for annotated games and finished games (not live games in progress): it calls \`handleExamineStart(gameDone)\` then \`handleExamineGoTo(idx)\`. The first examine of a done game resets to turn 0; the goTo is queued after it, so the clicked turn wins.
- The existing \`?turn=\` autocorrect effect picks up the new examined turn.

Tested with tsc, eslint and a unit test for the call order; not clicked through in a browser.


magratheazaphod:

Head: `57f9889`. **Verdict: ready to publish.** All files are product UI or tests. No secrets, and the commit message is clean.

**Correctness**
- **The clicked turn wins.** On the first examine of a finished game, `examineAndSeek` resets to turn 0 and then goes to the clicked turn in the same handler. React batches the two updates in order, and no effect overwrites the turn afterwards.
- **Live games are unchanged.** `gameDone` requires GAME_OVER, so moves stay unclickable during a game in progress.
- NIT: the first click on a finished game also switches on the analyzer's auto mode, the same as pressing Examine.

**Browser checks** (local stack, headless Chrome), clicking the play word:
- **Annotated game:** the first click without examining, clicks while examining, the last move, the two-column view, and iPhone taps all jump to the right position.
- **Board editor:** works, unchanged.
- **Regular game:** while in progress, nothing is clickable for players or viewers. Once finished, it works for both.
- `?turn=` counts events, not turns, so it can run ahead of "Turn N" in games with challenges or end-of-game points. That's existing behaviour.

**Checks:** `tsc`, lint and vitest (80 passed, 1 skipped) pass.

magratheazaphod:
Browser walkthrough of this branch (head `57f9889`), run locally: API, socket server and the rsbuild dev server from this branch, with a fresh database and one imported annotated GCG (a finished CSW15 game, 25 moves). Chrome, desktop layout.

1. Opened `/anno/<id>`. It loads at the final position (426-419), not in examine mode. All 25 scorecard moves carry the `main-word-seekable` class, i.e. are clickable (on master they have no `onSeek` here).
2. Clicked WHAC(K) (turn 23). The board jumped to the position before that move (WHACK not yet on the board, rack ACHW, 349), the move is highlighted, and examine mode and the analyzer opened. The clicked turn won over the reset to turn 0.
3. Scrolled the scorecard and clicked (U)H (turn 4). The board jumped to a board with only NULL on it, rack AGHINNS, and the URL became `?turn=5`.

One issue found: after the **first** click (step 2) the URL read `?turn=26` while the board showed turn 23. From the second click on it stays in sync. So on the first click the `?turn=` effect seems to run with the pre-jump turn count. Copying the link at that moment would open the final position, not the one shown.

magratheazaphod:
Correction to my walkthrough comment above: the "URL out of sync on first click" issue is not a bug. `?turn=` counts game events (`examinableGameContext.turns.length + 1`), while the scorecard header counts grouped turns. This game has 30 events, and WHAC(K) is event 26 but turn 23, so `?turn=26` was correct. I confirmed it in headless Chrome by logging the URL effect: after the click it computed 25 examined events, matching the jump. Arrow-key navigation on master uses the same numbering. Nothing to fix here; the walkthrough otherwise stands.

magratheazaphod:
More browser checks on head `57f9889` (local stack, headless Chrome), clicking the play word:
- Annotated game (`/anno`): first click without examining, second click while examining, last move, two-column view, and iPhone 13 emulated taps all jump to the right position.
- Board editor: unchanged, works.
- Regular game played locally by two users (6 passes, ended by consecutive zeroes), viewed at `/game/<id>`:
  - While the game was in progress, no move was seekable for either the player or a logged-out viewer.
  - After it ended, every row was seekable for both, and clicking rows 3 and 1 jumped to turns 3 and 1 (`?turn=3`, `?turn=1`).

magratheazaphod:
Real Jesse here: I tested this out and it works. The one thing I found a little surprising is that skipping to moves only works when clicking the word itself, as opposed to anywhere else in the cell but it looks like that was an intentional design decision by @dacronbb in https://github.com/woogles-io/liwords/pull/1826 so I won't touch that.

dacronbb:
Yeah, it's because sometimes there are other actions like editing a play or
leaving a comment in annotation mode, so it seemed like the simplest
decision to make only the text clickable.

magratheazaphod:
makes sense! I was wondering if it might impact mobile scrolling as well.

Issue #1712, opened 2026-03-04
Analyzer service does not work for annotated games


domino14:
analyzer service does not work for annotated games

Reported by @14domino via Discord

PR #1988, opened 2026-09-21
Fix opening annotated games via /game/<id>

magratheazaphod:
Opening an annotated game at `/game/<id>` shows "annotated game <id> should be accessed via GetDocument, not Get". This PR simply redirects to the corresponding '/anno/<id>' link.

Cause: since #1712, `GameStore.Get` rejects annotated games. `sendGameRefresher` (run from `initRealmInfo` on realm join) calls `Get` before its `Type == ANNOTATED` early return, so that check was unreachable and the error went back over the socket. The frontend redirect in `table.tsx` only fired when `GetMetadata` failed with that string, but `GetMetadata` succeeds for annotated games, so it never triggered.

Changes:
- `pkg/stores/game`: `Get` returns a wrapped sentinel `ErrAnnotatedGame` (same message text).
- `pkg/bus`: `sendGameRefresher` returns nil on `ErrAnnotatedGame`, replacing the dead type check. No extra DB query.
- `table.tsx`: when `GetMetadata` reports `GameType.ANNOTATED` and we're not in annotated mode, `location.replace` to `/anno/<id>`, keeping any query string and hash. Removed the error-string fallback, which can no longer trigger.

Other `/game/<id>` links (analysis-ready notification, profile game card/history, admin analysis stats, etc.) don't carry the game type, so they aren't changed; they now land on the redirect. Collections already route by `isAnnotated`.

Testing:
- `TestGetAnnotatedGame` (new): `Get` on an annotated game returns an error matching `ErrAnnotatedGame`. `pkg/stores/game` tests pass against local Postgres.
- Manual, local stack: `/game/<id>` for an annotated game showed the error without this fix and redirects to `/anno/<id>` on this branch.
- `tsc`, `eslint`, `vitest run src/gameroom` pass; `go build`/`go vet` clean.
- `sendGameRefresher` itself has no test; bus tests need more setup than it's worth here.


magratheazaphod:

Head: `455fc8b65bb2a6b729e7e8a3a042a3dfacc1bbd0`

**Verdict: PUBLISH AFTER FIXES** - one small should-fix.

### Files

| File | Call | Reason |
|---|---|---|
| `pkg/stores/game/db.go` | PUBLIC | Error sentinel for an existing check |
| `pkg/bus/gameplay.go` | PUBLIC | Skips the refresher for annotated games |
| `liwords-ui/src/gameroom/table.tsx` | PUBLIC | Redirects `/game/<id>` to `/anno/<id>` by game type |

### Findings

**SHOULD-FIX** - `pkg/stores/game/db.go:69-76`: `ErrAnnotatedGame` was inserted between `SetHistoryFetcher`'s doc comment and the function. The two comment blocks now merge onto the `var`, and `SetHistoryFetcher` loses its doc comment. Move the `var` (with its own comment) above the `SetHistoryFetcher` comment, or next to the other package-level declarations.

The fix itself is sound:
- `errors.Is` matches through the `%w` wrap.
- The message text is unchanged for any caller that still logs it.
- Returning nil only on this sentinel keeps every other `Get` failure surfacing as before.

The redirect now uses `GameInfoResponse.type`, which is reliable, where the old code matched the error string, which could no longer match. Links elsewhere that build `/game/<id>` for annotated games (analysis-ready notification, profile history, etc.) now reach the redirect. That fits, since those events don't carry the game type.

Test coverage: no test would have failed before the fix. The frontend redirect is page-load glue with no existing test pattern, and the bus tests need Postgres/Redis. The PR body says so.

### Checks run

- Secret / IP / path / email greps over the diff: no matches. Pass.
- `go build ./pkg/... ./cmd/...`: pass
- `go vet ./pkg/bus/ ./pkg/stores/game/`: pass
- `npx tsc --noEmit`: pass
- `npm run lint`: pass
- Not run: bus/store tests (need Postgres/Redis); a browser check of the redirect.

magratheazaphod:
Moved ErrAnnotatedGame (with its own comment) above SetHistoryFetcher so the func keeps its doc comment.

PR #1989, opened 2026-09-21
Exit from an annotated game returns to the editor

magratheazaphod:
From my experience, when a user is using the board editor, they often want to annotate multiple games in a row. Right now, finishing an annotated game requires going back to the main lobby and finding the board editor in the drop-down menu again, which is super-annoying.

You can still get back to the lobby simply by clicking EXIT TO EDITOR and then clicking the Woogles header.

If you're looking at someone else's annotated game it will still show BACK TO LOBBY. Maybe a better alternative in future would be BACK TO TOURNAMENT/CLUB/COLLECTION depending on how the user found the game.

-----
CLAUDE COMMENTS

The end-of-game Exit button calls \`handleExitToLobby\`, which navigated to the tournament slug or \`/\`. On an annotated game (\`/anno/:gameID\`) or in the board editor that dumped the annotator on the homepage.

\`BoardPanel\` now passes its existing \`annotated\` prop through to \`GameControls\`, and Exit goes to \`/editor\` when \`annotated\` or \`boardEditingMode\` is set. Other games are unchanged. I checked \`boardwizard/\` for other post-finish redirects to \`/\` and found none (the only navigate there goes to a broadcast after unclaiming).

Note: \`/editor\` resumes the user's latest unfinished annotation or starts a new one; there is no dedicated annotated-games list route to return to.

Tests added in \`game_controls.test.tsx\` for both destinations; tsc and eslint clean.


magratheazaphod:

Head: `2e8664c`. **Verdict: ready to publish.** All files are product UI or tests. No secrets, and the commit messages are clean.

**Behaviour.** For the annotated game's owner, both ways out go to `/editor`: the top-left link, which now reads "Back to editor", and the Exit button. Everyone else still sees "Back to lobby" and goes to the tournament page or `/`. The board editor always exits to `/editor`. Both controls share one `useOwnsAnnotatedGame` hook, so they use one rule and one cached `GetGameOwner` request.

**Addressed from review.** The first version sent every `/anno` viewer to `/editor`, which dropped strangers into their own editor and showed logged-out users an error toast. It is now limited to the owner, with tests for owner, non-owner and board-editor mode.

**Browser checks** (local stack, headless Chrome), using both the link and Exit:

| Viewer | Result |
| --- | --- |
| Owner | `/editor` |
| Another logged-in user | `/` |
| Logged out | `/` |

**Checks:** `tsc`, lint and the gameroom vitest suite (38 passed, 1 skipped) pass.

magratheazaphod:
Exit now goes to /editor only in board editing mode or when the current user owns the annotated game (same getGameOwner check as game_info.tsx); everyone else goes to the tournament or lobby. Tests cover owner, non-owner and board editing.

magratheazaphod:
Checked in headless Chrome against a local stack (head `4417035`), clicking Exit on a finished annotated game at `/anno/<id>`: logged out -> `/`, the game's owner -> `/editor`, a different logged-in user -> `/`. Matches the intended behaviour after the owner-only fix.

magratheazaphod:
Pushed `2e8664cdd`: the top-left link now reads "Back to editor" and goes to `/editor` for the annotated game's owner, matching Exit, so both ways out keep the annotator in the editing flow. Everyone else still sees "Back to lobby". The owner check moved into a shared `useOwnsAnnotatedGame` hook, so both controls use one rule and one cached `GetGameOwner` request. Checked in headless Chrome for the owner, a different logged-in user, and a logged-out viewer, using both the link and Exit.

PR #1990, opened 2026-09-21
Don't record auto-filled tiles as annotated event racks

magratheazaphod:
In annotated games the server tops both racks up to 7 with random tiles after every move, to keep tile accounting valid. When the annotator then entered a move without typing a rack, that random filler was saved as the event's rack:

- Exchange: the whole auto-dealt rack was recorded (e.g. dealt GMICEKG, exchanged GM, saved GMICEKG).
- Tile placement: if the played tiles happened to be in the random rack, the full random rack was recorded (dealt DOGXYZE, played D.G, saved DOGXYZE).

These bogus racks then flow into GCG exports and analysis.

Fix: `GameDocument` gets `known_racks`, the part of each rack the annotator entered or that a move implied. `AssignRacks` sets it from entered racks, played/exchanged tiles are removed from it, and it is clamped to the real rack after each move (tile borrowing etc.). The move is still made from the full rack, so tile conservation is unchanged, but the recorded event rack is the known tiles plus whatever the move itself shows. Replays (ImportGCG, amendments) rebuild it from event racks, so GCG racks behave as before. Documents without the field (native games, pre-existing annotated games) keep the old behaviour.

Tests added in `pkg/cwgame/api_test.go` for exchange and placement from an auto-filled rack, and for entered (full and partial) racks still being recorded as entered. `go test ./pkg/cwgame/` passes; the DB-backed omgwords tests were not run locally.


magratheazaphod:

Head: `e852027`. **Verdict: ready to publish.** It's an additive proto field plus regenerated code that matches it; no secrets, and the commit message is clean.

**Behaviour.** In annotated games, a move's recorded rack now holds only the tiles the annotator entered, plus the tiles the move itself showed. The server's random auto-fill is no longer saved. The new `GameDocument.known_racks` tracks those tiles, and only for annotated games. Native games and GCG imports are unchanged.

**Verified beyond the PR's tests.** In a scratch test, a phony played from an auto-filled rack was challenged off, then a pass and a full `ReplayEvents` followed. Tile accounting (`ReconcileAllTiles`) held after every step. The phony player kept only their known tiles, and the filler went back to the bag.

**Notes**
- Moves entered without a typed rack now record short racks, where they used to record fake 7-tile ones. That's the intent, but analysis will see short racks for those turns.
- #1991 also edits `omgwords.proto`. Whichever merges second must regenerate rather than hand-merge the generated files.
- NIT: `api/buf.gen.yaml` uses unpinned remote plugins, so every regeneration changes the version headers. Pinning them would be a small separate PR.

**Checks:** `go build`, `go vet` and `go test ./pkg/cwgame/` pass, as does `tsc`. The DB-backed tests weren't run.

magratheazaphod:
I validated that this works with manual annotated game creation. Entering exchanges from unknown racks remains annoying, but hopefully slightly easier with the spacebar rack entry shortcut from #1992

domino14:
how does this work for endgame situations, where racks are auto-assigned at the end and thus racks _are_ fully known for both players (if you enter one's rack)? can you look into it with Claude?

magratheazaphod:
Good catch - the endgame auto-assign fills racks only when they're unambiguous, so those racks were correct before this PR, and this change would have recorded them short. Fixed in 17578914e: once the bag is empty, if all racks but one are fully known, the last one is too. If neither rack was entered, the split stays unknown, since that's still just the random fill. Added a couple of tests for this case.

PR #1991, opened 2026-09-21
Time penalties in the board editor

magratheazaphod:
Annotated games could store and display time penalties (imported from GCG `(time)` lines) but there was no way to enter one in the editor.

- Proto: new `ClientGameplayEvent.TIME_PENALTY` (enum value 5) with `penalty_points` and `penalty_player_index`, since annotated games have no clocks and the annotator types the points.
- Server: a dedicated path in `ProcessGameplayEvent`, allowed only for annotated games that are already over (penalties come after end-of-game rack points, as in native games). Points must be 1-1000. It appends a `TIME_PENALTY` event with the player's rack, deducts the points, leaves the turn alone, and recomputes the winner.
- Replays: `ReplayEvents` and `ApplyEventInEditorMode` share the same apply logic, so the cumulative is recomputed from `LostScore` after an amendment and the winner stays correct. `handleEvent` now only reports the game as just ended on a real transition, so adding a penalty does not re-run the game-done hooks.
- ImportGCG: the replay part is pulled into `replayImportedHistory` so a GCG with a trailing `(time)` line can be tested without a DB.
- UI: when the game is over, the editor shows a "Time penalty" card (player, points defaulting to 10 with step 10) that sends the event via `sendGameEvent`.

Codegen: `buf generate` with the plugin versions pinned to the ones in the existing headers (protoc-gen-go v1.36.11, protoc-gen-es v2.13.0); only the omgwords files changed.

Tests: Go tests for applying a penalty, winner flip, replay round-trip, editor-mode recompute, rejections (not annotated, not over, bad points/player), and a GCG import with a trailing time penalty. Frontend test for the control.


magratheazaphod:

Head SHA reviewed: `b581be2f9b828c68fcdc9320fd2dc4e397a8acf0`

**Verdict: READY TO PUBLISH** - no blockers. One should-fix edge case and one merge-order note below.

### File classification

| File | Call | Reason |
| --- | --- | --- |
| api/proto/ipc/omgwords.proto | PUBLIC | new enum value + two fields, additive |
| rpc/api/proto/ipc/omgwords.pb.go | PUBLIC | regenerated, matches proto |
| liwords-ui/src/gen/api/proto/ipc/omgwords_pb.ts | PUBLIC | regenerated, matches proto |
| pkg/cwgame/api.go | PUBLIC | penalty processing + replay/editor handling |
| pkg/cwgame/api_test.go | PUBLIC | tests |
| pkg/omgwords/handlers.go | PUBLIC | justEnded only on transition to GAME_OVER |
| pkg/omgwords/service.go | PUBLIC | ImportGCG refactor into replayImportedHistory() |
| pkg/omgwords/import_test.go | PUBLIC | test |
| pkg/omgwords/testdata/vs_josh_time_penalty.gcg | PUBLIC | test fixture, public-figure nicknames only |
| liwords-ui/src/boardwizard/editor.tsx | PUBLIC | editor wiring |
| liwords-ui/src/boardwizard/time_penalty.tsx | PUBLIC | new UI card |
| liwords-ui/src/boardwizard/time_penalty.test.tsx | PUBLIC | test |

Secret / IP / home-path / email greps over the diff: no hits. Single commit, message is clean.

### Findings

**SHOULD-FIX - `pkg/cwgame/api.go:711` (ApplyEventInEditorMode) / `api.go:766` (applyTimePenalty): no play-state guard on re-apply.**
`processTimePenalty` enforces GAME_OVER, but the editor-mode re-apply path does not. When an amendment to an earlier move leaves the re-applied game not over (e.g. the going-out play is amended so it no longer goes out), `handleAmendment` re-applies the saved TIME_PENALTY via `ApplyEventInEditorMode` and it succeeds, leaving a penalty event in the middle of a live game with further moves appendable after it. Fix: in the `GameEvent_TIME_PENALTY` case of `ApplyEventInEditorMode`, return an error when `gdoc.PlayState != GAME_OVER`, so the existing truncate-on-failure logic drops it (the annotator can re-enter it once the game ends again). A one-line test in `api_test.go` would pin it.

**NOTE - merge order with #1990.** #1990 also edits `api/proto/ipc/omgwords.proto` and regenerates the same two files (`rpc/api/proto/ipc/omgwords.pb.go`, `liwords-ui/src/gen/api/proto/ipc/omgwords_pb.ts`) for a different message (`GameDocument.known_racks = 25`). The serialized descriptor (Go rawDesc / TS `fileDesc(...)` base64) is one blob, so whichever PR merges second must rebase and re-run `gen.sh` rather than textually resolve the conflict.

Verified, no finding:
- **handleEvent `justEnded` change (`handlers.go:49,104`)**: on the non-amendment path, `ProcessGameplayEvent` returns `errGameNotActive` for every event type other than TIME_PENALTY when the game is already over, so `wasOver == true` can only reach the check for a time penalty. Normal endings (PLAYING -> GAME_OVER) still set `gameEnded`, so `MarkAnnotatedGameDone` / `onGameDone` fire exactly as before. Amendments return early through `handleAmendment`, which is untouched and still reports `GAME_OVER` as ended (existing behaviour for re-ending amendments preserved). Net effect: adding a penalty does not re-trigger the done hooks, which is the intent.
- **Cumulative recompute from LostScore**: `MacondoEvtToOMGEvt` copies `LostScore` (pkg/entity/utilities/document.go), so previously imported GCG penalties replay to the same cumulative as before; events with `LostScore == 0` fall back to the stored cumulative (old default-branch behaviour). Native games never go through `cwgame.ReplayEvents`/`ApplyEventInEditorMode` (their penalties come from `pkg/gameplay/end.go` on the macondo path), so they are unaffected.
- **Authorization parity**: the new event goes through the existing `SendGameEvent`, which does `AuthUser` + `GameOwnedBy` like every other editor event; `processTimePenalty` additionally requires `GameType_ANNOTATED`, GAME_OVER, `1..1000` points and a valid player index, and the rejection test confirms the document is untouched on error.
- **Generated code**: decoded the TS `fileDesc` descriptor and compared it with the Go package's registered descriptor: identical apart from the dependency list the TS generator omits by design. Field numbers 7/8, enum value 5 and comments match the proto; no version-header churn.
- ImportGCG refactor is a pure extraction; done-marking and `UpdateDocument` order unchanged.

### Checks run

- `go build ./pkg/... ./cmd/... ./rpc/...` - pass (`go build ./...` at repo root fails for an unrelated pre-existing reason).
- `go vet ./pkg/cwgame/... ./pkg/omgwords/...` - one warning, `handlers.go` "assignment copies lock value" at `*g = *gdocClone`; present on master (line 233), not introduced here.
- `go test ./pkg/cwgame/...` - pass (includes the new TestTimePenalty, TestTimePenaltyEditorModeRecomputesCumulative, TestTimePenaltyRejected).
- `go test ./pkg/omgwords/ -run TestReplayImportedHistoryTimePenalty` - pass. Other DB-backed omgwords tests not run (no Postgres available).
- `npx tsc --noEmit` - pass; `npx eslint` on the three boardwizard files - pass; `npx prettier --check` - pass.
- `npx vitest run src/boardwizard/time_penalty.test.tsx` - 2/2 pass.
- Generated-descriptor equality check (TS vs Go) - equal as described above. `buf` is not installed here, so a full regenerate-and-diff was not run.

magratheazaphod:
ApplyEventInEditorMode now errors on a saved TIME_PENALTY when the game isn't over, so the existing truncate-on-failure drops it instead of applying it mid-game. ReplayEvents is unchanged; added a test.

magratheazaphod:
This looks atrocious right now, I'll go back to the drawing board on how to present this. Maybe a modal.

